PR #33, opened 2015-03-01
Demo build configuration with gradle

stiXits:
I wrote a little gradle configuration to generate the plugin zip's, in case you're interested :)


decebals:
Thanks

PR #39, opened 2015-05-27
Fix build on JDK 1.8

lightglitch:


pom.xml
```diff
@@ -112,6 +112,15 @@
 
     <profiles>
         <profile>
+            <id>JDK 8 Build</id>
```
decebals: I prefer an id without space characters (for example jdk8-build)

lightglitch:
Updated.

decebals:
Thanks

Issue #32, opened 2015-02-28
Support plugins dependency version validation


lightglitch:

decebals:
@lightglitch Now, with #44 we can do this. There is interest for this function?

lightglitch:
I'm interested but I'm a little busy this week. As soon as I have time I will look into it.

decebals:
It will be resolved by #150.

decebals:
Closed by #150.

PR #44, opened 2015-06-05
Replace Version with semver lib

lightglitch:
First Step to #32


decebals:
My philosophy (see my projects, [pippo](https://github.com/decebals/pippo) for example) is to keep the number of dependencies as low as possible, because people/users want this. I add a new dependency only if that library brings value in project.
I think that Version is a decent, uncompleted implementation of semver.  If you see some problems in this class please tell me.

lightglitch:
I understand that and the value that the lib brings is the support for semver ranges (https://github.com/zafarkhaja/jsemver#semver-expressions-api-ranges) that I'm going to need for compatibility check of plugins.

decebals:
Thanks for your explanation. Can you comment on how you see # 32 implementation to have a common understanding?

lightglitch:
I still thinking about it because I want to use the process information of the graph to do the validations.

But as a big picture I just going to put the plugin with the disable state if some of his dependencies are disabled. This with some error reporting information is enough for me.

decebals:
@gitblit I think that it's a good idea to have range validation on plugin version and I think that the `jsemver` library is a solid choice (no dependency, active development - I don't wish to implement a complex semver in pf4j). Is it ok for you this change?

gitblit:
It's ok with me.

pf4j/src/main/java/ro/fortsoft/pf4j/PluginDependency.java
```diff
@@ -12,38 +12,38 @@
...
-
-	@Override
-	public String toString() {
-		return "PluginDependency [pluginId=" + pluginId + ", pluginVersion=" + pluginVersion + "]";
-	}
+    private String pluginId;
+    private String pluginVersionSupport;
+
+    public PluginDependency(String dependency) {
+        int index = dependency.indexOf('@');
```
decebals: why `@` and not `:`?

lightglitch: For me the sintax plugin-at-version makes more sense but I saw that you were using : and forgot to change it.

decebals: No problem, we can keep `@`

PR #59, opened 2015-07-21
A few more unit tests

lightglitch:


pf4j/pom.xml
```diff
@@ -43,6 +43,12 @@
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
```
decebals: :+1: I like mockito.

pf4j/src/main/java/ro/fortsoft/pf4j/PluginWrapper.java
```diff
@@ -54,7 +54,7 @@ public String getPluginPath() {
 	 * for this plug-in. The class loader can be used to directly access
 	 * plug-in resources and classes.
 	 */
-    public PluginClassLoader getPluginClassLoader() {
+    public ClassLoader getPluginClassLoader() {
```
decebals: I know that this change makes `PluginWrapper` testable but I think that it's a drawback from the API point of view. If it's not another option and if this change doesn't break the current code (at compilation) I am ok with this modification.

pf4j/src/test/java/ro/fortsoft/pf4j/DefaultExtensionFactoryTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ *
+ * @author Mario Franco
+ */
+public class DefaultExtensionFactoryTest {
+
+    public DefaultExtensionFactoryTest() {
```
decebals: I think that we can remove empty constructor and empty `setUp`, `tearDown` methods to strip the code. Is it ok for you?

pf4j/src/main/java/ro/fortsoft/pf4j/PluginWrapper.java
lightglitch: It doesn't break the current code (you can check the travis build of the pull request), another option is to also mock the PluginClassLoader in the test.

pf4j/src/test/java/ro/fortsoft/pf4j/DefaultExtensionFactoryTest.java
lightglitch: Sure.

pf4j/src/main/java/ro/fortsoft/pf4j/PluginWrapper.java
lightglitch: Another thing the PluginClassLoader doesn't expose any extra API only the ClassLoader public methods so I don't see a big drawback here.

decebals: I am ok with this modification.

lightglitch:
@decebals Clean up is done.

decebals:
:+1: Thanks!

Issue #53, opened 2015-07-13
Thread safety in PluginClassLoader


antonywarner:
Hi, we are starting to experiment with PF4J (thanks for the work by the way) and have come across a bug where the implementation of loadClass in PluginClassLoader is not thread safe. When we invoke it concurrently it tries to load the same class twice resulting in a LinkageError:

java.lang.LinkageError: loader (instance of ro/fortsoft/pf4j/PluginClassLoader): attempted duplicate class definition for name: "com/foo/MyExtension"
at java.lang.ClassLoader.defineClass1(Native Method) ~[na:1.7.0_80]

I notice that java.lang.ClassLoader#loadClass is made thread safe with a per class name lock:

protected Class<?> loadClass(String name, boolean resolve)
        throws ClassNotFoundException
    {
    synchronized (getClassLoadingLock(name)) {
        ...
    }
}

PluginClassLoader overrides that method and so loses that synchronized block. I think the fix is as simple as adding in a synchronized block? The getClassLoadingLock method is protected and so available to PluginClassLoader.

For now we have synchronised access to DefaultPluginManager.

decebals:
I am in vacation. I will come back Monday.

decebals:
Thanks for reporting this issue. The request for make pf4j thread safe appears also in #30. To be honest, I am not a specialist in java concurrency but I will try to do something in this direction.

decebals:
I pushed a fix so you can test your application with the latest SNAPSHOT (now is 0.10.0-SNAPSHOT) via the Sonatype Maven Repository:

```
<repositories>
    <repository>
        <id>sonatype-nexus-snapshots</id>
        <url>https://oss.sonatype.org/content/repositories/snapshots</url>
        <releases>
            <enabled>false</enabled>
        </releases>
        <snapshots>
            <enabled>true</enabled>
        </snapshots>
    </repository>
</repositories>
```

PR #62, opened 2015-08-06
Feature/custom class loaders

mark-gresham:
We are using pf4j as part of our application, but we needed to modify the behaviour of `PluginClassLoader` to verify the classes being loaded. In order to achieve this I introduced a new interface `PluginClassLoaderFactory` which was used as a API extension to select which type of `ClassLoader` to use for a plugin. Our alternate `ClassLoader` uses `PluginClassLoader` as a base.
To ensure that the existing behaviour remains, `DefaultPluginClassLoaderFactory` was implemented.
Altering `ClassLoader`s should probably be considered an _advanced_ use of the framework, something which most users wouldn't need to implement.

### DIFF
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginClassLoaderFactory.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginClassLoaderFactory.java
@@ -1,0 +1,10 @@
+package ro.fortsoft.pf4j;
+
+public class DefaultPluginClassLoaderFactory implements PluginClassLoaderFactory {
+  
+    @Override
+    public PluginClassLoader getPluginClassLoader(final PluginManager pluginManager,
+            final PluginDescriptor pluginDescriptor, final ClassLoader parent) {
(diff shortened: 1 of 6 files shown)


decebals:
Please check #63 (a much simple implementation) to see if satisfy your requirements. Thanks!

mark-gresham:
The changes introduced in #63 look like they could be sufficient for our needs.

mark-gresham:
When could you do a release with #63 and #53 included?

decebals:
See https://github.com/decebals/pf4j/issues/21#issuecomment-129364073. I can release a new version today if the SNAPSHOT is ok for you.

mark-gresham:
Apologies for the delay in replying. We have ran the snapshot code against our codebase and everything is working. If you haven't already done it, could you please do the release?

decebals:
PF4J 0.10.0 released, see https://groups.google.com/forum/#!topic/pf4j/onwDMxheAcA. The new version will be available soon on Maven Central Repository.

PR #66, opened 2015-09-08
added possibility to overwrite DefaultPluginManager (to create f.i. a JarPluginManager)

tuxedo0801:
To create an "JarPluginManager" that uses JARs (f.i. created with Maven Assembly "jar-with-dependencies") instead of ZIP, I needed to add two get-methods to DefaultPluginManager.

And while adding those two get-methods, I also updated javadoc to keep maven javadoc generation happy.

Here is my JarPluginManager implementation incl. demo program:

https://gist.github.com/tuxedo0801/0bc4ad265131b2d46643

Instead of the long configuration part in plugin's pom,xml + assembly xml, you now only need this in the pom (just an example):

....
    <properties>
    
        <plugin.id>knx-logic-plugin</plugin.id>
        <plugin.class>de.root1.kad.logicplugin.LogicPlugin</plugin.class>
        <plugin.version>1.0.0</plugin.version>
        <plugin.provider>AC</plugin.provider>
        <plugin.dependencies></plugin.dependencies>    
    </properties>
....
    <build>
        <plugins>
            <plugin>
                <artifactId>maven-assembly-plugin</artifactId>
                <version>2.5.4</version>
                <configuration>
                    <archive>
                        <manifestEntries>
                            <Plugin-Id>${plugin.id}</Plugin-Id>
                            <Plugin-Class>${plugin.class}</Plugin-Class>
                            <Plugin-Version>${plugin.version}</Plugin-Version>
                            <Plugin-Provider>${plugin.provider}</Plugin-Provider>
                            <Plugin-Dependencies>${plugin.dependencies}</Plugin-Dependencies>
                        </manifestEntries>
                    </archive>
                    <descriptorRefs>
                        <descriptorRef>jar-with-dependencies</descriptorRef>
                    </descriptorRefs>
                </configuration>
                
                <executions>
                    <execution>
                        <id>make-assembly</id>
                        <phase>package</phase>
                        <goals>
                            <goal>single</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>	            
        </plugins>
    </build>
....

That creates an JAR archive with "-jar-with-dependencies.jar" suffix/extension which includes all classes from dependencies.

One could also think of skipping the plugin-extraction-process and just pass the JAR to the classloader. But for improved startup-speed on small systems (Raspberry Pi f.i.) it could be a good idea to extract the JAR.

### DIFF
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginManager.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginManager.java
@@ -131,6 +131,23 @@ public class DefaultPluginManager implements PluginManager {
         initialize();
     }
 
+    /**
+     * Returns the currently used direcotry containing the plugins
+     * @return File pointing to the plugins directory
+     */
+    public File getPluginsDirectory() {
+        return pluginsDirectory;
+    }
+
+    /**
+     * Returns the PluginClasspath used by this plugin manager
+     * @return  the classpath used by this manager
+     */
+    protected PluginClasspath getPluginClasspath() {
+        return pluginClasspath;
+    }
+    
+
     @Override
     public void setSystemVersion(Version version) {
     	systemVersion = version;
@@ -689,7 +706,7 @@ public class DefaultPluginManager implements PluginManager {
      * If getRuntimeMode() returns RuntimeMode.DEVELOPMENT than a
 	 * DEVELOPMENT_PLUGINS_DIRECTORY ("../plugins") is returned else this method returns
 	 * DEFAULT_PLUGINS_DIRECTORY ("plugins").
-     * @return
+     * @return File pointing to the plugins directory
      */
     protected File createPluginsDirectory() {
     	String pluginsDir = System.getProperty("pf4j.pluginsDir");
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultExtensionFactory.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultExtensionFactory.java
@@ -16,8 +16,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * The default implementation for ExtensionFactory.
- * It uses Class.newInstance() method.
+ * The default implementation for ExtensionFactory. It uses Class.newInstance()
+ * method.
  *
  * @author Decebal Suiu
  */
@@ -26,9 +26,11 @@ public class DefaultExtensionFactory implements ExtensionFactory {
     private static final Logger log = LoggerFactory.getLogger(DefaultExtensionFactory.class);
 
     /**
-     * Creates an extension instance. If an error occurs than that error is logged and the method returns null.
+     * Creates an extension instance. If an error occurs than that error is
+     * logged and the method returns null.
+     *
      * @param extensionClass
-     * @return
+     * @return an extension instance
      */
     @Override
     public Object create(Class<?> extensionClass) {
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginFactory.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginFactory.java
@@ -31,7 +31,7 @@ public class DefaultPluginFactory implements PluginFactory {
     /**
      * Creates a plugin instance. If an error occurs than that error is logged and the method returns null.
      * @param pluginWrapper
-     * @return
+     * @return a plugin instance
      */
     @Override
     public Plugin create(final PluginWrapper pluginWrapper) {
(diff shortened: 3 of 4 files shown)


pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginManager.java
```diff
@@ -131,6 +131,23 @@ public DefaultPluginManager(File pluginsDirectory) {
...
+     */
+    public File getPluginsDirectory() {
+        return pluginsDirectory;
+    }
+
+    /**
+     * Returns the PluginClasspath used by this plugin manager
+     * @return  the classpath used by this manager
+     */
+    protected PluginClasspath getPluginClasspath() {
```
decebals: I don't think that we need this method. From your gist code I see that you use this method in 
```java
protected PluginDescriptorFinder createPluginDescriptorFinder() {
    if (RuntimeMode.DEVELOPMENT.equals(getRuntimeMode())) {
        return new PropertiesPluginDescriptorFinder();
    }

    return new DefaultPluginDescriptorFinder(getPluginClasspath());
}
```
but your code is exactly the same with the default implementation:
```java
protected PluginDescriptorFinder createPluginDescriptorFinder() {
    if (RuntimeMode.DEVELOPMENT.equals(getRuntimeMode())) {
        return new PropertiesPluginDescriptorFinder();
    }

    return new DefaultPluginDescriptorFinder(pluginClasspath);
}
```

pf4j/src/main/java/ro/fortsoft/pf4j/DefaultPluginManager.java
```diff
@@ -131,6 +131,23 @@ public DefaultPluginManager(File pluginsDirectory) {
         initialize();
     }
 
+    /**
+     * Returns the currently used direcotry containing the plugins
+     * @return File pointing to the plugins directory
+     */
+    public File getPluginsDirectory() {
```
decebals: In your scenario if I "relaxed" `pluginsDirectory` variable from "private" to "protected" I think that we resolve the problem.

decebals:
Other comments for your gist.

`JarPluginClasspath` implementation is to verbose. A short version could be: 
```java
public class JarPluginClasspath extends PluginClasspath {

    public JarPluginClasspath() {
        super();
    }

    protected void addResources() {
        classesDirectories.add(".");
    }

}
```

decebals:
To be honest I don't see the value for a big jar plugin (jar with dependencies) versus a zip plugin (with the default layout: `classes` and `lib`) but this is other story.

tuxedo0801:
You're absolutely right. It's enough if pluginsDirectory is protected. That would be sufficient for overwriting the DefaultPluginManager.

The method "getPluginClasspath()" is also not required. It's a try&error-relict while trying to hack the JarPluginManager. Sorry for the confusion.

I updated the gist sample files + added a partial pom.xml for a plugin.

> To be honest I don't see the value for a big jar plugin (jar with dependencies) versus a zip plugin (with the default layout: classes and lib) but this is other story.

The advantage is:

* No additional assembly.xml per plugin, less xml in pom.xml
* One can depend in parent/distribution maven project on the plugin artifact like this:

```
        <dependency>
            <groupId>de.root1.kad</groupId>
            <artifactId>kad-logic-plugin</artifactId>
            <version>1.0.0-SNAPSHOT</version>;.cdelmpru
            <classifier>jar-with-dependencies</classifier>
        </dependency>
```

which enables you to use the plugin, f.i. in another assembly (which packages the main application plus required packages into a distribution zip) like this:

assembly.xml
```
.....
        <dependencySet>
            <includes>
                <include>de.root1.kad:kad-logic-plugin::jar-with-dependencies</include>
            </includes>
            <outputDirectory>plugins</outputDirectory>
            <useProjectAttachments>true</useProjectAttachments>
        </dependencySet>
.....
```

With that, I'm now in the position of specifiying the version of my main-application plus the factory-default-plugins with its version in the pom of the "distribution" project and bundle all the stuff together to big distribution zip.

Maybe there are other ways of handling .zip artifacts of projects and bundle them together with maven assembly plugin. But that's the default way the maven documentation describes.


Beside that: The ZIP files worked well. I don't see the requirement of having the JarPluginManager in PF4J project. But there should be the option to overwrite the DefaultPluginManager to create a custom manager.

decebals:
Probably the best option is to extend `DefaultPluginManager` (are a lot of stuff that can be configured via `createXYZ` factory methods. To implement a `PluginManeger` from zero I think that it's a hard work.

decebals:
If you need more help please tell me. Don't forget to close this PR if everything is OK.

tuxedo0801:
> Probably the best option is to extend DefaultPluginManager (are a lot of stuff that can be configured via createXYZ factory methods.)

That was also my feeling. Maybe it's also a good idea to set a few other fields to "protected" instead of private? Otherwise overwriting other createXYZ methods could result in the same issue...

For me and my "special case" it's okay now. Thanks for changing the code.

decebals:
> That was also my feeling. Maybe it's also a good idea to set a few other fields to "protected" instead of private? Otherwise overwriting other createXYZ methods could result in the same issue...

Sure, I will do this step by step (on request).

tuxedo0801:
I'm working on an JarPluginManager which does not need to extract the jar archive. Will see if this works and what else needs to be changed in visibility.

PR #67, opened 2015-09-10
Add automatic application and plugins assembly using maven

lightglitch:
Added to the example demo project


lightglitch:
I added the pf4jplugin module (used as a plugin in maven assembly) to the project but can be moved to the demo if anyone think is better to be there.

decebals:
I spent few minutes to understand how it works, it's a little unclear for me but probably because I am not a maven expert. I will try again and I will come with a conclusion. Thanks!

lightglitch:
It was based on this article http://dacanal.blogspot.pt/2014/12/how-to-assemble-multi-module-project.html

decebals:
I will evaluate again this PR.

PR #68, opened 2015-09-10
PF4J site for github pages

lightglitch:


decebals:
Wow. Super. How can I see the result? How can I modify the content? 
I have a something similar for another [project](https://github.com/decebals/pippo). The source of the site is available [here](https://github.com/decebals/pippo-site). I used Jekyll.

lightglitch:
Hi, It's based on the https://github.com/lviggiano/owner site http://owner.aeonbits.org/ which is almost the same as http://jekyllrb.com/ 

by running antrun:run it will run jekyll locally and you can see it on localhost.

decebals:
For the moment, we will go with readme file and probably with wiki from Github. 
The argument is the lack of time on my part

PR #87, opened 2016-01-18
Fix issue "URI is not hierarchical" when working with extensions jar's

horuzhiy:
Replace working with content of plugin jars with NIO.
new File(url.toURI()) -> FileSystems.newFileSystem(url.toURI() ...


decebals:
Thanks!

PR #89, opened 2016-02-12
Fix "URI is not hierarchical" issue

harbulot:
Pull request #87 fixed this problem for `readClasspathStorages()` but not for `readPluginsStorages()`. This applies the same fix.


decebals:
Thanks!

PR #94, opened 2016-02-23
Used plugin2 as a example project on how to handly 3rd party libraries with gradle

lyze237:
Libraries like org.json or something like that would be an example.
You basically need to pack them into: <name>.zip/lib/<library>.jar

Whenever you open the created zip in a file manager it may or may not display the "lib" directory properly but it's there. All those file managers can't display a path like: "<name>.zip/classes/../jar/<library>.jar" but the plugin manager can.

Also fixed a little copy paste error in plugin2/build.gradle. We don't work in the "welcome" package but in the "hello" package there.

### DIFF
--- a/demo_gradle/.gitignore
+++ b/demo_gradle/.gitignore
@@ -1,2 +1,4 @@
 build
 .gradle
+
+app/plugins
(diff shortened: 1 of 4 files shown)


decebals:
Please create another PR because this is unmergeable.

PR #95, opened 2016-02-23
Added Legacy Plugin Manager.

lyze237:
Added the `LegacyPluginManager` which basically is a override from the `DefaultPluginManager` but uses the `LegacyExtensionFinder` instead of the `DefaultPluginManager`


decebals:
I think that is obsolete because the actual implementation already contains this code.

Issue #103, opened 2016-06-30
Gradle demos don't build zip with libs


ecumene-software:
In the gradle demos, they don't build the plugin with the library dependencies.
`demo_gradle/plugins/plugin1/build.gradle`
`demo_gradle/plugins/plugin2/build.gradle`

A simple change in the build script could change this, I've got the solution in my app if you'd like me to create a pull request?

Or you can do it yourself in an upcoming commit:
From:
```
task plugin(type: Jar) {
    baseName = 'WelcomePlugin'
    version = '0.1.0'
    into('classes')
    extension('zip')
    with jar
}
```
To:
```
task plugin(type: Jar) {
    baseName = 'WelcomePlugin'
    version = '0.1.0'
    into('classes'){
         with jar
    }
    into('lib'){
        from configurations.compile
    }
    extension('zip')
}
assemble.dependsOn plugin // This is just so it builds the plugin with the build task
```
The code above should create two folders, lib and classes, in the plugin's zip. **This has a little issue where it copies the API as a jar into lib** but that can be fixed by adding a new configuration group called extraLibs

decebals:
The Gradle was submitted by a contributor. I don't use Gradle in my projects so please submit a PR if something is wrong. I don't know if we need a readme file (in demo_gradle folder) to explain about how to build and run the demo with Gradle.
Thanks.

PR #104, opened 2016-06-30
Fix for issue #103 and added the SLF4 Simple dependency

ecumene-software:
Fix for issue #103 and added the SLF4J simple to the gradle demo's app project


decebals:
Thanks for your contribution!

PR #111, opened 2016-09-14
Add inheritance support on Extension annot

elecomte-pro:
Hi @decebals, here a small update for inheritance support on extension model. No need for AnnotationProcessor changes. Works well in my project.


decebals:
Thanks!

PR #122, opened 2017-02-16
Updated pf4j to newest version.

lyze237:
Fixed compilation errors after update.
Added copyPlugins task to gradle which copies all generated plugin.zip's to the app/plugins directory.


decebals:
Thanks! 
I don't use Gradle in my project, so I cannot validate this PR but I trust that it's OK.

Issue #123, opened 2017-02-16
Can't find extensions when api uses a different package name then "ro.fortsoft.pf4j.demo.api"


lyze237:
Hello!

I've checked out the demo_gradle project of the code base and tried to add a couple extensions.

Whenever I add a new ExtensionPoint, lets say "com.thyllian.Hallo" with the following content:

    package ro.fortsoft.pf4j.demo.api;

    import ro.fortsoft.pf4j.ExtensionPoint;

    public interface Hallo extends ExtensionPoint {
        public String getHallo();
    }

And add the newly created ExtensionPoint to "plugin1/WelcomeGreeting":

    @Extension
    public static class WelcomeGreeting implements Greeting, Hallo {
    	@Override
        public String getGreeting() {
            return "Welcome";
        }
        @Override
        public String getHallo() {
            return "Hallo";
        }
    }

Pf4j can't find WelcomeGreeting when executing pluginManager.getExtensions(Hallo.class);

Now however when I move the "at.lyze.api.Hallo" class to "ro.fortsoft.pf4j.demo.api.Hallo" and fix all imports it it works.

Log files:
* ro.fortsoft.pf4j.demo.api: https://paste.ubuntu.com/24007186/
* at.lyze.api: https://paste.ubuntu.com/24007188/

Log Diff: https://www.diffchecker.com/mDmlDaJw

MANIFEST.MF will get created properly in both cases:

    # Generated by PF4J
    ro.fortsoft.pf4j.demo.welcome.WelcomePlugin$WelcomeGreeting
    ro.fortsoft.pf4j.demo.welcome.WelcomePlugin$WelcomeGreeting

Project Files: 
* at.lyze: https://transfer.sh/2FDX9/at.lyze.zip
* ro.fortsoft.pf4j.demo: https://transfer.sh/jsvWo/ro.fortsoft.pf4j.demo.zip

decebals:
First of all I want to say that it's probably one of the best issue report  which I have seen. :+1: 
It's for the first time then I see in PF4J that someone tries to create an extension class (annotated with `@Extension`) that implements multiple extension points. An issue is that we shouldn't have three entries/lines in `extensions.idx` (for `plugin2`) with same content.
Can you test the same approach with `Hallo` in `at.lyze.api` package but with three separate extensions in plugin2 (one extension class for each extension point)?

lyze237:
I have the following setup now:

api: at.lyze.api.Hallo:

[...]

PR #124, opened 2017-02-16
Added compileOnly to "api" dependencies in plugins.

lyze237:
Fixed issue: #123 

Same as: https://github.com/decebals/pf4j/blob/master/demo/api/pom.xml#L32


decebals:
Thanks!

PR #125, opened 2017-02-17
fix npe

ShaoboAi:
findClassNames may return null


decebals:
Can you supply more information (in what scenario)?
Also please attach in PR only the modification without to reformat the whole file.
Thanks!

decebals:
Thanks!

PR #129, opened 2017-03-30
Properties Descriptor finder bug fixes and a test

janhoy:
Adding parsing of plugin.description and plugin.requires to the properties finder, and added a test


pf4j/src/test/java/ro/fortsoft/pf4j/PropertiesPluginDescriptorFinderTest.java
```diff
@@ -0,0 +1,177 @@
+package ro.fortsoft.pf4j;
```
decebals: Please add the copyright header (copy/paste from other .java file).

decebals:
Thanks!

PR #130, opened 2017-03-30
Refactor validation of PluginDescriptors

janhoy:
Was duplicated in each finder, now pluggable.
Did this through an AbstractPluginDescriptorFinder so everyone get the default behavior ootb.
To plug in your own validator you'd need to Override `createPluginDescriptorFinder()` in your PluginManager.


decebals:
My idea is to move the validation process from `PluginDescriptorFinder` to `PluginManager`, after we found a PluginDescriptor (https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/ro/fortsoft/pf4j/AbstractPluginManager.java#L753).
The simple solution is to create a protected method `validatePluginDescriptor` in AbstractPluginManager and to call this method for each PluginDescripotor instance (no more PluginDescriptorValidator).
What you say?

janhoy:
Agree that this was a more elegant location for the validation. I also had to fix some of the validation tests that previously failed by explicitly calling validation from the test code.
Alternatively we could have created a separate test class for validation of incomplete descriptors.

decebals:
I propose you to move the code from PluginDescriptorUtils in AbstractPluginManager and to delete:
- PluginDescriptorUtils class
- code that tests validity of PluginDescriptor from *PluginDescriptorFinderTest.java (revert the modifications)

You can add a `testValidatePluginDescriptor` in `AbstractPluginManagerTest`. In the future if we see that validatePluginDescriptor method is a candidate for an extension point (many overwrites for this method) we can create an interface PluginDescriptorValidator with a default implementation (DefaultPluginDescriptorValidator) but in this moment I think that it's not necessary to do this (we complicate the library and increase artificially the learning curve ). 

Is it OK for you?

janhoy:
Think I've accommodated most of it now. Found no AbstractPluginManagerTest so I created a DefaultPluginManagerTest

decebals:
Thanks! Good work.

janhoy:
Good solution, easy to plug own validation. In my scenario I accept a descriptor without a PluginClass since Solr already has another way of handling that.

PR #131, opened 2017-03-30
Fix bug in loadJars()

janhoy:
DefaultPluginLoader loads my plugin just fine, it manages to add `classes/` to the PluginClassLoader, but the jars in `lib/` are not added.

Turn out that `loadJars()` refuses to add jars if the folder File object is not an absolute path.

Also adds `slf4j-simple` logger in test scope and convert some File usage to nio Path.


pf4j/src/test/java/ro/fortsoft/pf4j/PluginLoadingTest.java
```diff
@@ -0,0 +1,21 @@
...
+import static junit.framework.TestCase.assertNotNull;
+
+/**
+ * Test loading of plugins
+ */
+public class PluginLoadingTest {
+    @Test
+    public void testPluginLoader() throws ClassNotFoundException {
+        PluginManager manager = new DefaultPluginManager();
+        assertNotNull(manager.loadPlugin(Paths.get("src/test/resources/plugin-with-jar")));
```
decebals: Is this file available only on your machine?

pf4j/src/main/java/ro/fortsoft/pf4j/util/FileUtils.java
```diff
@@ -96,27 +92,27 @@ public static boolean delete(File fileOrFolder) {
 		return success;
 	}
 
-	public static List<File> getJars(File folder) {
+	public static List<File> getJars(Path folder) {
```
decebals: I am OK with the modifications from this class. My intention is to move this code to Java 7 NIO API, and to use `Files.newDirectoryStream` (less code lines and probably faster).

pf4j/pom.xml
```diff
@@ -33,6 +33,12 @@
             <version>1.7.5</version>
         </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
```
decebals: Do we need this? In all my libraries projects I use slf4j-api as dependency and I use what slf4j I wish in application.

janhoy: This added dependency is an slf4j logger implementation with a **test** scope so it will allow tests to output logs. The runtime jars will NOT contain this class - as you say, the user app decides what to use.

janhoy:
Wow, test coverage from 17% to 32% :-)

decebals:
Please remove `PluginLoaderTest` class from this PR because comes with .class files in resources and I don't like this. Maybe we will find other method to test PluginLoader (for example via mock object).

decebals:
Thanks! I prefer small PR that solves only one problem.

PR #133, opened 2017-03-30
Support for adding license information to the plugins

janhoy:
e.g. Apache-2.0 or MIT
Also adds missing vars `description` and `requires` to toString


decebals:
Thanks!

PR #134, opened 2017-04-03
getVersion() use wrong class for calculating PF4J version

janhoy:
I have a custom PluginManager, and when the `getVersion()` code attempts to log PF4J's version in initialize(), it uses `getClass()` to fetch implementationVersion. This assumes that getClass returns a PF4J class, but it returns my own class, which incidentially has a version not parsable by semVer Version class (it contains a space and git hash after the version string).

The fix is to always use a pf4j class for calculating the version.


decebals:
Thanks!

PR #135, opened 2017-04-03
deletePlugin() failed to delete plugin folder with contents

janhoy:
`deletePlugin(id)` which is called by `UpdateManager.uninstallPlugin(id)` failed since the `PluginRepository.deletePluginPath()` call used `Files.deleteIfExists(path)` which is not able to delete recursively.

The fix is to use the already existing `FileUtils.delete(path)`, which I modified to `java.nio` style and no follow symlinks.


decebals:
Thanks!

PR #136, opened 2017-04-03
Delete plugin zip on uninstall

janhoy:
Delete plugin zip file from `pluginsRoot` as part of `deletePlugin(id)`

### DIFF
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/util/FileUtils.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/util/FileUtils.java
@@ -118,4 +118,33 @@ public class FileUtils {
             }
         }
     }
+
+    /**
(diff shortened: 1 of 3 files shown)


pf4j/src/main/java/ro/fortsoft/pf4j/util/FileUtils.java
```diff
@@ -118,4 +118,33 @@ private static void getJars(final List<File> bucket, Path folder) {
...
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Delete a file (not recursively) and ignore any errors
+     * @param path the path to delete
+     */
+    public static void optimisiticDelete(Path path) {
```
decebals: Wrong spelling for `optimistic` ?

decebals:
Thanks!

PR #137, opened 2017-04-03
The requires Expression does not print well

janhoy:
Instead we get the className. As I could not find any method on Expression to convert back to String, this fix adds a `requiresString` variable to PluginInfo.


decebals:
Make sense. Thanks!

PR #138, opened 2017-04-04
Refactor of requires in PluginDescriptor (breaking change)

janhoy:
Also added CHANGELOG entries


decebals:
Excellent work. Thanks!

PR #139, opened 2017-04-05
Ability to get pluginsRoot from PluginManager

janhoy:
Simple fix which will make https://github.com/decebals/pf4j-update/pull/11 more elegant


decebals:
Thanks!

PR #140, opened 2017-04-05
Unzip plugin zip file in loadPluginFromPath()

### LINKED ISSUE #132: Missing expand archive step in version 1.2.0
In version 1.1.0, the `DefaultPluginManeger.loadPlugin` method expands the plugin archive. In version 1.2.0 (big refactoring to offer an improved customization), this step is missing.
The https://github.com/decebals/pf4j-update/issues/4 issue is an result of this inconvenient.

janhoy:
This is an attempt for auto unzip on load, so UpdateManager can install and update without problems. Was this the approach you had in mind?

Fixes #132

### DIFF
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/AbstractPluginManager.java
@@ -18,7 +18,6 @@ package ro.fortsoft.pf4j;
 import com.github.zafarkhaja.semver.Version;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import ro.fortsoft.pf4j.util.FileUtils;
 import ro.fortsoft.pf4j.util.StringUtils;
 
 import java.io.Closeable;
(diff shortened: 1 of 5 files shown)


decebals:
The code looks good with one observation on my part. 
The part with plugin packed as a zip file and the unpack process before plugin loading, is available in `Default` implementation (`DefaultPluginManager`, `DefaultRepository`, ...), the `AbstractPluginManager` knows nothing about this aspect. From this reason I think that it's a good idea to move the expand zip line from `AbstractPluginManager.loadPluginFromPath` to `DefaultPluginManager.loadPluginFromPath`.

```java
// pseudo code
public class DefaultPluginManager extends AbstractPluginManager {
 
    ... other methods

    Override
    protected PluginWrapper loadPluginFromPath(Path pluginPath) throws PluginException {
        // first unzip any ZIP files
        try {
            pluginPath = FileUtils.expandIfZip(pluginPath);
        } catch (Exception e) {
            log.warn("Failed to unzip " + pluginPath, e);
            return null;
        }

        return super.loadPluginFromPath(pluginPath); // where pluginPath is the value returned by expandIfZip
    }

}
```

Did what I said make sense to you?

decebals:
Excellent. Thanks!

PR #141, opened 2017-04-06
New LoadPluginsTest

janhoy:
Also:
Test-logging using log4j
Bugfix for expandIfZip - would fail on new folders without "." in the name
Bugfix for duplicate check (pathToIdMap)

The new test should give a huge bump in test coverage :-)

### DIFF
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/util/FileUtils.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/util/FileUtils.java
@@ -166,11 +166,12 @@ public class FileUtils {
      * @throws IOException on error
      */
     public static Path expandIfZip(Path filePath) throws IOException {
-        String fileName = filePath.getFileName().toString();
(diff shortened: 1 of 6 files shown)


pf4j/src/main/java/ro/fortsoft/pf4j/util/FileUtils.java
```diff
@@ -167,7 +167,7 @@ public static void optimisticDelete(Path path) {
      */
     public static Path expandIfZip(Path filePath) throws IOException {
         String fileName = filePath.getFileName().toString();
```
decebals: Is it better/clear something like above snippet?

```java
public static boolean isZipFile(Path path) {
    if (Files.isDirectory(path)) { // or !Files.isRegularFile(path)
        return false;
    }

    return path.toString().toLowerCase().endsWith(".zip");
}
```

decebals:
The code looks good. I don't have to much time to analyse the unit tests but I feel that they are OK. I wait your response on my comment (a possible little improvement in FileUtils) and after this, I will merge it.

decebals:
Thanks!

PR #144, opened 2017-04-19
Refactor requires validation, Fixes #142

### LINKED ISSUE #142: Interpretation of "requires" when exactly x.y.z
Today the `PluginDescriptor` and pf4j-update's `PluginInfo` takes a `requires` in string format and converts it into an expression when enforcing, using `ExpressionParser.parse()`.

My question/suggestion is in the case where developer does not enter an expression like `2.x` or `1.0 - 3.1` but enters an exact x.y.z version e.g. `2.3.1`, should that be taken literally so the plugin will only start with exactly that version, or should we rewrite that corner case into meaning `>=2.3.1` or less extreme `^2.3.1` (which is effectively 2.3.1-2.4.0)? 

Projects should of course educate plugin writers. But perhaps the framework could benefit from some kind of global configuration option e.g. `interpretExactRequiresAsMinVersion` defaulting to false?

janhoy:
* Introduce config flag exactVersionAllowed to interpret exact versions as >= expression
* Moves Expression parsing and validation from PluginDescriptor to PluginManager
* Added tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,10 +256,14 @@ a requirement to specify a minimum system version for loading plugins.
 Loading & starting a newer plugin on an older system could result in
 runtime failures due to method signature changes or other class
 differences.  
+
 For this reason was added a manifest attribute (in PluginDescriptor) to specify a 'requires' version
-which is a minimum system version. Also DefaultPluginManager contains a method to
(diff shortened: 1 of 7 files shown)


decebals:
Thanks!

PR #149, opened 2017-06-19
Updated gradle demo dependencies and switched from S.out.println to slf4j

lyze237:
Updated pf4j and commons gradle dependencies to newest version and changed all build settings so we can use slf4j in plugins#148 

we needed to exclude all slf4j libraries from all plugins to get them to work.


decebals:
Thanks! 
I don't use Gradle in my projects so I cannot help on this subject.

PR #154, opened 2017-07-05
Refactor ManifestPluginDescriptorFinder to allow custom managing of V…

tcollignon:
Hello,

This is a small PR to allow custom managing of version.
In my case version may be : 1.1.1.RC1, and https://github.com/zafarkhaja/jsemver will not handle this, and throw exception.

Thanks

### DIFF
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/ManifestPluginDescriptorFinder.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/ManifestPluginDescriptorFinder.java
@@ -76,8 +76,15 @@ public abstract class ManifestPluginDescriptorFinder implements PluginDescriptor
         return pluginDescriptor;
     }
 
+    /**
+     * Parse version to semver {@link Version} object.
+     * Example : 1.1.1.RC1 -> 1.1.1-RC1.
+     * This may be override to use a custom parsing to semver Version.
+     * @param version given string to be the plugin version
+     * @return a semver version
+     */
     protected Version createPluginVersion(String version) {
-      return Version.valueOf(version);
+        return Version.valueOf(version);
     }
 
     protected PluginDescriptor createPluginDescriptorInstance() {
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/ManifestPluginDescriptorFinder.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/ManifestPluginDescriptorFinder.java
@@ -58,7 +58,7 @@ public abstract class ManifestPluginDescriptorFinder implements PluginDescriptor
 
         String version = attributes.getValue("Plugin-Version");
         if (StringUtils.isNotEmpty(version)) {
-            pluginDescriptor.setPluginVersion(Version.valueOf(version));
+            pluginDescriptor.setPluginVersion(createPluginVersion(version));
         }
 
         String provider = attributes.getValue("Plugin-Provider");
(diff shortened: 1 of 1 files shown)


janhoy:
So what `Version` object would you return when parsing `1.1.1.RC1`? Are you not in control of the versioning of the plugin so that you could comply with SemVer?

tcollignon:
Hi,

I return Version object, but before send it,  I convert last "." to "-" in string version

janhoy:
Ah, now I see, the RC1 part is valid SemVer, as long as it is separated with a dash.
But again, isn't it easier to put the correct syntax in the manifest to start with, if you're in control of the build :)

tcollignon:
Build part is big and I prefer keep my versionnig as is :)
I think java-semver is not felxibe enough, it raise too much exception :(

decebals:
@tcollignon 
>I think java-semver is not felxibe enough, it raise too much exception :(

[Semantic versioning](http://semver.org/) is a specification, and [jsemver](https://github.com/zafarkhaja/jsemver) is a possible implementation in Java, used by PF4J. In the first versions, PF4J used a plain [Version](https://github.com/decebals/pf4j/blob/release-0.9.0/pf4j/src/main/java/ro/fortsoft/pf4j/Version.java) class but I was forced to adopt jsemver because it comes with nice support for range and comparisons. I think that was a correct decision.

> In my case version may be : 1.1.1.RC1 ...

Definitely it's not a valid version according SemVer. Your PR is an workaround and I feel that it's not enough. For example, in #150 I implemented something like:
```java
/**
 * Check if an existing version of dependency is compatible with the required version (from plugin descriptor).
 *
 * @param requiredVersion
 * @param existingVersion
 * @return
 */
protected boolean checkDependencyVersion(String requiredVersion, Version existingVersion) {
    return existingVersion.satisfies(requiredVersion);
}
```
where `requiredVersion` is the string version specified as dependency in the plugin descriptor.
If you declare "plugin-1" with version "1.1.1.RC1" and "plugin-2" has a dependency to "plugin-1@1.1.1.RC1" (a non semver) you are in trouble. Probably they are other places where we force a semver.

Now, I trying to see if it's possible to interconnect PF4J witj JMPS (Java 9 modules).  The project is in initial phase (https://github.com/decebals/pf4j-jpms). I gathering information about JMPS () and try to collect specifications about a possible solution. One possible problem is that JMPS doesn't embrace semver. To accommodate this, in the future it is possible to add an abstraction `VersionManager` with a default implementation that works with semver (like current implementation). If someone want to use a custom version concept, he must implement a new version manager and inject that manager in `PluginManager`.

So, bottom line, I feel that the best/solid approach for you would be to go with semver and to convert your plugins versions to this type of version.
If you feel that this PR is valuable for you, please add some javadoc to the new method `createPluginVersion` (add motivation, your usecase with the conversion 1.1.1.RC1 -> 1.1.1-RC1), correct the indentation and I am ready to appli it :smile: .

tcollignon:
Hi,

And thanks for your good explanation, I understand what you mean and what you tring to do. I just tell that actually is a bit complicated to deal with jsemver with all kind of versioning. So I try to convert my versioning to semver when I run plugin with pf4J (and only for that).

I think a VersionManager will be a very good thing for your project

So I will fix my PR with your recommandations

Tx again

decebals:
@tcollignon Can you try the variant with `VersionManager` (https://github.com/decebals/pf4j/pull/155). I updated the SNAPSHOT version.

tcollignon:
thanks for this work, it seem great !
I don't know if I can test this right now, but I add this in my stack

PR #155, opened 2017-07-11
Add VersionManager abstractization (breaking change)

decebals:
See https://github.com/decebals/pf4j/pull/154 for motivation.

### DIFF
--- a/pf4j/src/main/java/ro/fortsoft/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/ro/fortsoft/pf4j/AbstractPluginManager.java
@@ -15,7 +15,6 @@
  */
 package ro.fortsoft.pf4j;
 
-import com.github.zafarkhaja.semver.Version;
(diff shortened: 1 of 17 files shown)


decebals:
This PR comes with a `VersionManager` interface:
```java
/**
 * Manager responsible for versions of plugins.
 *
 * @author Decebal Suiu
 */
public interface VersionManager {

    /**
     * Check if a {@code constraint} and a {@code version} match.
     *
     * @param constraint
     * @param version
     * @return
     */
    boolean satisfies(String constraint, String version);

}
```

The default implementation is `DefaultVersionManager` and it uses SemVer. You are not forced to use SemVer in your projects. If you want something custom, then override `createVersionManager` method from `AbstractPluginManager` and supply your implementation.

**Note**  This PR comes with a breaking change.

janhoy (approved):
Probably a good change. Did a quick look only.

PR #170, opened 2017-09-21
A demo with fewer/simpler maven configs

thammegowda:
## Background
I was giving this project a test run to see if I can use this for [one of my project](https://github.com/USCDataScience/sparkler). I am very impressed; I am planning to swap this in the place of OSGI! __Thank you for developing this and making it available under Apache Licence 2.0.__
I looked into the demo app and felt that the maven configs can be a bit better and simpler.
So I went ahead and changed it (in the way I am planning to use in my project). and thought of submitting back.

## About this  PR
This is a derived work of `../demo` project with following changes:

+ The pom.xml is unbelievably simpler. No more complicated maven-plugin configurations.
+ No other XMLs except `pom.xml`. Not even `assembly.xml`
+ Uses plugin.properties, also the plugin versions are synchronized with maven artifact version (which will be updated with maven release plugin if configured).
+ Everything is a .jar and the application runs without extracting the content.
  Trying to avoid .zip and extractions since it requires write permissions at the runtime which may be unavailable to users (especially where many users share the same installation of the application).

PS: if you like to replace existing demo with this new demo, I will be happy to make those changes.

### DIFF
--- a/demo2/README.md
+++ b/demo2/README.md
@@ -1,0 +1,14 @@
+# Demo 2 Application
+
+This is an application to demonstrate the use of pf4j.
+This is a derived work of `../demo` project with following changes:
+
++ The pom.xml is unbelievingly simpler. No more complicated maven-plugin configurations.
++ No other XMLs except `pom.xml`. Not even `assembly.xml`
++ Uses plugin.properties
++ Everything is a .jar and the application runs without extracting the content.
+  Trying to avoid .zip and extractions since it requires write permissions at the runtime which maybe unavailable
+  to some users (especially many users share the same installation of application).
+
+
+Run `./run-demo.sh` from this directory for a demo.
--- a/demo2/api/pom.xml
+++ b/demo2/api/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <parent>
+        <groupId>org.pf4j.demo</groupId>
+        <artifactId>pf4j-demo-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>pf4j-demo-api</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+    <name>Demo Api</name>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-deploy-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.pf4j</groupId>
+            <artifactId>pf4j</artifactId>
+            <version>${pf4j.version}</version>
+            <!-- !!! VERY IMPORTANT -->
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/demo2/api/src/main/java/org/pf4j/demo/api/Greeting.java
+++ b/demo2/api/src/main/java/org/pf4j/demo/api/Greeting.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2012 Decebal Suiu
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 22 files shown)


decebals:
Thanks for your appreciations. 
My opinion is that your demo treats only a particular case (a version for plugins distributed as jar files). I will try to extract some ideas from your PR and to propose an improved version of the current demo (from maven point of view).

decebals:
Possible directions to improve the maven support:
 - create an maven archetype similar with one from my project https://github.com/decebals/pippo/tree/master/maven-archetypes/quickstart (http://www.pippo.ro/dev/quickstart.html) 
- write a custom Maven plugin

thammegowda:
> I will try to extract some ideas from your PR and to propose an improved version of the current demo 

👍  Maven should be simple to use if we follow convention over configuration, right?

+1 for maven archetype!

decebals:
From Pippo I understood the importance of quickstart (based on a Maven archetype). I think that it's good to add the same concept in PF4J. 
I feel that the Maven archetype should be based on your work from this PR, because the approach used by you is the simplest and the use case is the most common (plugin as a fat .jar file). 
What I don't want to use is the zero configuration in case of `development` mode. Now, in development mode if I use the structure from demo (with `api`, `app`, `plugins` and each plugin is a module in plugins) I can run the application (with all plugins) from my IDE (Idea IntelliJ in my case) like a normal application. 
No extra configuration, just specify the development mode with `-Dpf4j.mode=development`.
To accommodate your maven scripts with actual development mode I think that it's enough to merge the code from your `JarPropsPluginDescriptorFinder` in `PropertiesPluginDescriptorFinder` (if `plugin.properties` is not found in the root of the plugin path try to see if I can find that file in the classpath of plugin).

thammegowda:
> I feel that the Maven archetype should be based on your work from this PR, because the approach used by you is the simplest and the use case is the most common (plugin as a fat .jar file).

👍  👍 

> What I don't want to lose is the zero configuration in development mode.

Yes. I agree. I didnt test it in development mode. In fact, development mode is MUST have feature. Please tweak this to support development mode. 

Regarding `JarPropsPluginDescriptorFinder`
Actually, I just wanted to use  plugin's classloader and ask `pluginClassLoader.getResourceAsStream("plugin.properties")`. That should take care of resolving the path - whether the properties file is in `src/main/resources` in development mode or in the jar file/extracted directory of deployment mode. 
Is plugin classloader available during the time of accessing properties file?

decebals:
> Is plugin classloader available during the time of accessing properties file?

You have right and the current implementation is correct, first we read the plugin descriptor from the plugin path and after this, if everything is OK (plugin descriptor contains all needed information) we load the plugin.

image

Your implementation is correct.
The next step is to apply your code (https://github.com/thammegowda/pf4j/blob/281496c130fa510a07393bc3fdb396c603676121/demo2/app/src/main/java/org/pf4j/demo/Boot.java#L136) to `PropertiesPluginDescriptorFinder`, build and install locally the new pf4j version and test both modes (development and deployment). If everything is OK we can apply code on origin/master and create the maven-archetypes/quickstart module.

thammegowda:
Hey @decebals 
The code is updated as per your suggestions.
But one test case is failing - `org.pf4j.PropertiesPluginDescriptorFinderTest.testFindNotFound`
I know exactly why it is failing - 
When the loader cant find plugin's properties, it raises `org.pf4j.PluginException` but the test case expects `java.nio.file.ProviderNotFoundException`. Which exception is appropriate here?

thammegowda:
The fix is applied to the main code base and the demo is simplified. All tests Passed 💯

thammegowda:
The fix is applied to the main code base and the demo is simplified. All tests Passed 💯

decebals:
@thammegowda Sorry for delay. I will try to implement the concept describe in https://github.com/decebals/pf4j/issues/143#issuecomment-332331245. Our approach described here has some weak points:
- in [JarPluginManager](https://github.com/decebals/pf4j/blob/5916ee64d9c05a4f085f6f6562ed383ff3203e4a/pf4j/src/main/java/org/pf4j/JarPluginManager.java#L48), `PropertiesPluginDeccriptorFinder` is used only in dev mode
- I don't want to loose `ManifestPluginDescriptorFinder` (it's extended by JarPluginDescriptorFinder), some people maybe use `JarPluginManager` already (no backwards compatibility)

So, my idea is to play a little with the "compound" concept (as it's described in https://github.com/decebals/pf4j/issues/143#issuecomment-332331245) in a separate branch.
I will come with news.

decebals:
See https://github.com/decebals/pf4j/pull/172 and the branch [compound](https://github.com/decebals/pf4j/tree/compound).

decebals:
Finished the implementation of `CompoundPluginDescriptorFinder`. The concept is powerful and it's a model/template for other components (`CompoundPluginLoader`, `CompoundPluginRepository`).
An updated SNAPSHOT is already available.
So, I will close this PR and the code with demo will be transformed in a Maven archetype available as a new module in `maven-archetypes/quickstart` directory.

PR #172, opened 2017-09-27
Add CompoundPluginDescriptorFinder

decebals:
A POC (proof of concept) for https://github.com/decebals/pf4j/issues/143#issuecomment-332331245.


janhoy:
Looks nice. I can give more feedback once I try to move from my homegrown solution to this one. But that won't happen until next week or so.

PR #175, opened 2017-10-11
Improve public API of Extention* classes

styurin:
`ExtensionWrapper` and `ExtensionDescriptor` have poor public API which doesn't allow using them outside of `org.pf4j` package, but these classes are parts of public API of pf4j library.

This change converts `ExtensionDescriptor` to be an immutable object with all public fields that cannot be changed after creation. Also, removed `setExtensionFactory` from `ExtensionWrapper` in favor of a constructor parameter.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/ExtensionDescriptor.java
+++ b/pf4j/src/main/java/org/pf4j/ExtensionDescriptor.java
@@ -20,23 +20,12 @@ package org.pf4j;
  */
 public class ExtensionDescriptor {
 
-    private int ordinal;
-    private Class<?> extensionClass;
+    public final int ordinal;
+    public final Class<?> extensionClass;
 
-    public Class<?> getExtensionClass() {
-        return extensionClass;
-    }
-
-    public int getOrdinal() {
-        return ordinal;
-    }
-
-    void setExtensionClass(Class<?> extensionClass) {
-        this.extensionClass = extensionClass;
-    }
-
-    void setOrdinal(int ordinal) {
+    public ExtensionDescriptor(int ordinal, Class<?> extensionClass) {
         this.ordinal = ordinal;
+        this.extensionClass = extensionClass;
     }
 
 }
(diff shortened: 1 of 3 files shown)


pf4j/src/main/java/org/pf4j/ExtensionDescriptor.java
```diff
@@ -20,23 +20,12 @@
  */
 public class ExtensionDescriptor {
 
-    private int ordinal;
-    private Class<?> extensionClass;
+    public final int ordinal;
```
decebals: I have a curiosity. Why you make the fields public and remove the accessors?
I don't say that is wrong but I am curious to  know your motivation.

decebals:
Thanks! Good job.
In the same situation is [PluginDescriptor](https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/PluginDescriptor.java). For `PublicDescriptor` I think that the best approach is to use a builder (builder pattern) because we have many fields. I have this refactoring in pending from a long time but it has never been a priority.

pf4j/src/main/java/org/pf4j/ExtensionDescriptor.java
styurin: Because it's an immutable class, its variables cannot be changed after creation. It's also very simple class for POJOs which is used to combine multiple other objects and doesn't have other logic which depends on these variables.

decebals: I know what it is an immutable class and the advantages of such construction.
My question is described in details on https://softwareengineering.stackexchange.com/questions/190955/using-public-final-rather-than-private-getters.
Thanks for your explanations.

josiahhaswell:
@decebals Thanks for the update.  My feeling is that builders are fantastic for very public facing APIs, but clunky for semi-internal APIs like *Descriptor.  If that's the direction you want to go, I can convert the PR to use builders.  However, this is blocking adoption of PF4J for me, so if we could figure it out ASAP that would be wonderful.

PR #176, opened 2017-10-13
Emit warning instead of error when plugin root is not present

styurin:
The framework can properly function without plugins root by loading plugins from classpath so there is no reason to log an error, warning should be enough.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -208,7 +208,7 @@ public abstract class AbstractPluginManager implements PluginManager {
         log.debug("Lookup plugins in '{}'", pluginsRoot);
         // check for plugins root
         if (Files.notExists(pluginsRoot) || !Files.isDirectory(pluginsRoot)) {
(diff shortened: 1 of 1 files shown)


decebals:
> The framework can properly function without plugins root so there is no reason to log an error, warning should be enough.

Can you elaborate this statement please?

styurin:
> Can you elaborate this statement please?

Plugins can be loaded from classpath which means plugins root is not strictly necessary.

decebals:
> ...  which means plugins root is not strictly necessary.

It's an idea to move this constrain (existence of plugins root directories - variable and method) in `DefaultPluginManager` to make `AbstractPluginManager` more abstract. 

> Plugins can be loaded from classpath ...

I think that a such approach is a step back. It's a possibility to load each plugin with the same `ClassLoader` (return the same class loader from `PluginLoader.loadPlugin` method).
In the end it's not clear for me the use case. If the plugins are in the classpath and maybe plugins are jars, then  I don't know why you want to use PF4J.

It's interesting to know your use case.

styurin:
> It's interesting to know your use case.

It's useful during migration when everything is in one jar and then moving individual plugins to separate jars.

PR #177, opened 2017-10-21
ISSUE-169: fix gradle demo

acesar:
* Fixed Gradle demo
* Updated PF4J version for the demo
* Added a README that explains the demo


decebals:
Thanks!

PR #178, opened 2017-10-26
@Override should not change method signature

kkriske:
when subclassing `DefaultPluginManager` you are forced to create a `CompoundPluginDescriptorFinder` even if you only need one


decebals:
Initial, my intention was to change the return value for all methods that return internally a `Compound` class implementation. The idea was to avoid a cast in subclass. The construction is OK from Java point of view.
In the end I think that it's good to preserve the original method signature without to alter the return value.

decebals:
Thanks!

PR #197, opened 2018-01-22
close JarFile stream after getManifest

HeartUnchange:
According to the [Plugin lifecycle](https://github.com/decebals/pf4j#plugin-lifecycle), a plugin stopped should be able to be removed. But it couldn't. Threre still is some stream opened.

One is in the PluginDescriptor, the other is the URLClassLoader.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/ManifestPluginDescriptorFinder.java
+++ b/pf4j/src/main/java/org/pf4j/ManifestPluginDescriptorFinder.java
@@ -51,8 +51,8 @@ public class ManifestPluginDescriptorFinder implements PluginDescriptorFinder {
 
     protected Manifest readManifest(Path pluginPath) throws PluginException {
         if (FileUtils.isJarFile(pluginPath)) {
-            try {
-                Manifest manifest = new JarFile(pluginPath.toFile()).getManifest();
+            try(JarFile jar = new JarFile(pluginPath.toFile())) {
+                Manifest manifest = jar.getManifest();
                 if (manifest != null) {
                     return manifest;
                 }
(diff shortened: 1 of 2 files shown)


pf4j/src/main/java/org/pf4j/ManifestPluginDescriptorFinder.java
```diff
@@ -51,8 +51,8 @@ public PluginDescriptor find(Path pluginPath) throws PluginException {
 
     protected Manifest readManifest(Path pluginPath) throws PluginException {
         if (FileUtils.isJarFile(pluginPath)) {
-            try {
-                Manifest manifest = new JarFile(pluginPath.toFile()).getManifest();
+            try(JarFile jar = new JarFile(pluginPath.toFile())) {
```
decebals: Good catch!

decebals:
@HeartUnchange 
The fix for `ManifestPluginDescriptorFinder` is good.
The second fix, for `AbstractPluginManager.stopPlugins()` I think that it's not good. You can delete a plugin only after you unload the plugin (https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/PluginManager.java#L106). 
In `unloadPlugin` method (https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L281) we close the class loader.
The idea is that after a stop plugin you can call start plugin and the same class loader is used.
Actually, it's a `deletePlugin` method (https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/PluginManager.java#L130) that delete a plugin from system (inclusive the jar file).

I think that it's a good idea to submit the fix for `ManifestPluginDescriptorFinder` from a separate PR (pull request) and I will merge it.
After this we can comment here, on the second submit, if you believe that your contribution for `stopPlugins` is OK.

HeartUnchange:
I am sorry. I didn't notice the [unloadPlugin](https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L281) method. Just change ```ManifestPluginDescriptorFinder``` now. About the ```stopPlugins``` , I will create a new issue.

decebals:
Thanks!

Issue #203, opened 2018-03-09
File lock on plugin jar not released


nonnamon:
If a plugin is misconfigured e.g. incorrect Plugin-class name in the manifest file, it is still able to be 'loaded' and 'resolved' but unable to be started or unloaded. 

Is it possible to release the Jar file handle since the plugin cannot be started, such that the file can be over-written? If not then the plugin jar file will be kept open by the process and cannot be overwritten on the file system until process exit.

Conceptually, since the plugin has been 'loaded' it should be possible to 'unload' it too? Currently this fails during attempt to stopPlugin(id)

decebals:
>If a plugin is misconfigured e.g. incorrect Plugin-class name in the manifest file, it is still able to be 'loaded' and 'resolved' 

It's not so clear for me if it's a bug in PF4J or not.
Probably if a plugin class doesn't exist then that plugin should me marked as in error.
All the aspects about the Plugin lifecycle are described in http://www.pf4j.org/doc/plugin-lifecycle.html.

If a plugin is resolved means that all the plugins that depend on the plugin in question can move forward. A plugins uses classes from dependencies and probably this is the reason why the jar is locked.

Do you receive any error? I am curios to know how PF4J signals this issue.

nonnamon:
from memory, a class not found exception happens out of the class loader when attempting to instantiate the class using the incorrect name in the misconfigured manifest. I will try to reproduce with demo and provide some logs.

i agree if the class does not exist should be marked as error, this doesn't seem to be the case currently as the plugin gets to 'resolved'. if a plugin is unable to be used should be able to overwrite the plugin file without restarting the host process.

decebals:
> i agree if the class does not exist should be marked as error, this doesn't seem to be the case currently as the plugin gets to 'resolved'.

Plugin resolved means that jar (storage) exists and the dependencies also and the plugin descriptor is valid (all mandatory fields exist).
If the plugin class is missing from a plugin (jar) then that plugin should be in error and not in resolved state. I think that here is the problem (bug).
I will investigate the case.

nonnamon:
[issue203.txt](https://github.com/decebals/pf4j/files/1799267/issue203.txt)

Logs attached.

e.g. in demo metadat.mf, add required Plugin-xxx fields, use incorrect name 'WelcomePluginX' for Plugin-class. Plugin reaches resolved:

```
[...]

PR #204, opened 2018-03-16
plugin verification info should be keeped

hxzhao527:
If a plugin disabled by [plugin verification](https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L738), the ```DISABLED``` state shoud be keeped.


HeartUnchange:
Although this change can solve issue #203 , but there is not a proper state for it according to the plugin lifestyle now.

hxzhao527:
Closed. Because #223 have bring this in.

PR #210, opened 2018-04-05
Fixes #206

### LINKED ISSUE #206: Support multiple plugin directories
Currently, only a single plugins directory is supported as the pluginRoot. This directory is set to `../plugins` in the development mode `plugins` in the production mode. The `DefaultPluginManager` also accepts the pluginRoot path as an argument but default to the above if not specified. It would be good to support multiple plugins directories as pluginRoot. The primary purpose would be to segregate different types of plugins into different directory locations.

shirishpandharikar:
Removed unnecessary call to set pf4j.pluginsDir if plugin root is set via the constructor.


decebals:
Thanks!

Issue #195, opened 2018-01-18
Gradle demo not working with recent intellij IDEA


anidotnet:
I tried the gradle demo with the latest IntelliJ IDEA (2017.3.3) and it is not working anymore. They have recently made some fundamental changes in gradle plugin and it is restricting the gradle project to run properly. It is not finding the extension.idx anymore. Happening for both Java and Kotlin projects. Maven demo are working fine though.

Do you have any insight on this how to avert this?

decebals:
@anidotnet 
Personal, I don't know how can I help in this issue (I don't know/use Gradle).
I feel that you wrote a comment in the past to JetBrains related to this subject (maybe an old version).
Maybe other users that use PF4J and Gradle can help.

anidotnet:
Yes, for reference here is the issue I raised with JB - https://youtrack.jetbrains.com/issue/KT-19823
But unfortunately, pf4j still does not work though they claims they have fixed it. I am desperately looking for a solution to this problem.

decebals:
Why you need the last IDEA? Is it not an option to use an old,  tested, IDEA?

anidotnet:
Yeah, that is also an option, but again I'll miss out on the new features introduced by the IDE :)

shirishpandharikar:
I tried the Gradle demo and it does not work either with Eclipse or Idea. Looking at the Pf4j code DevelopmentPluginClasspath adds the target/classes and target/lib to the plugin classpath. So when the plugin classloader is constructed the plugin classes do not end up within the classloader. My plugins are getting resolved but fail to start with a ClassNotFound exception. Am I doing something wrong here?

With Gradle, the output directories can be customized easily. I tried doing so by adding the following to the plugins/build.gradle file within the subprojects task project.buildDir = 'target'
This fixes the issue of the plugin class not being found. But when I package the plugins using Gradle I end up having 2 extensions.idx files in the META-INF directory.

Any pointers here would be really helpful.

decebals:
@shirishpandharikar Can you show us the PF4J's log?

decebals:
@shirishpandharikar Maybe https://github.com/decebals/pf4j/issues/192 can help.

shirishpandharikar:
Here are the logs from my run:
[main] INFO org.pf4j.demo.Boot - ########################################
[main] INFO org.pf4j.demo.Boot -                PF4J-DEMO                
[main] INFO org.pf4j.demo.Boot - ########################################
[...]

PR #221, opened 2018-06-01
Improving development mode experience for Gradle users #195

ssherwood:
Added additional filter for "build" folder which is the default output folder for Gradle projects.


decebals:
Thanks!

Issue #207, opened 2018-04-02
List all extension points and extensions for a plugin


decebals:
I think that it's nice to have a possibility to list all extension points and extensions that come with a plugin. 
It's nice to have a kind of plugin's documentation created automatically.

decebals:
We have already this information (see [LegacyExtensionStorage](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/processor/LegacyExtensionStorage.java#L73)), all we have to do is write this information in the `extensions.idx` file.

decebals:
The first step in my opinion is to improve the format (content) of the `extensions.idx` file.
Now the format is very simple, with one extension id (class) on each line. As example I will present you the content of `extensions.idx` for the `welcome` plugin (see demo):
```
# Generated by PF4J
org.pf4j.demo.welcome.WelcomePlugin$WelcomeGreeting
``` 

My propose is to add also on the same line, the extension point(s) implemented by extension, separated by  `:` as delimiter:
```
# Generated by PF4J
org.pf4j.demo.welcome.WelcomePlugin$WelcomeGreeting:org.pf4j.demo.api.Greeting
```

So, we have the extension (`org.pf4j.demo.welcome.WelcomePlugin$WelcomeGreeting`) and the extension point (`org.pf4j.demo.api.Greeting`) separated by `:` delimiter character.
The first big benefit is that we can avoid the reflection (time consuming) at runtime to discover the extension point(s) for an extension.
The second benefit is that we can create a nice documentation.

decebals:
To list all the extension points defined by a plugin, I think that we need a separate (probably) storage  `extension-points.idx`. The format of the new file can be identical with the actual format of the `extensions.idx`, with one extension point id (class) on each line.
Other idea is to put this information (with all extension points) in the already existed `extensions.idx`.

decebals:
I went through the code again today, so here is where this stands.

First a clarification about the title. Listing the extensions of a plugin has been possible since before this issue was opened, with `getExtensionClassNames(String pluginId)`. What is missing is everything about extension points: which extension point each extension implements, and which extension points a plugin defines. Neither is available today.

[...]

PR #233, opened 2018-07-24
Allow a way to query all extension classes for a given plugin

dustContributor:
What the title says. Was looking for that method in the API but I didn't see it, was there a particular reason why is it missing?

In any case, it'd work exactly the same as the other methods since the default finder is already capable of doing it.

I am partial to creating an array and returning with Arrays.asList instead of creating a new ArrayList (since as an object it's a bit more compact, or just returning an array is viable too), but that'd be a different behavior from the rest of the methods.


decebals:
Thanks!
It's an important pass forward for implementation of https://github.com/pf4j/pf4j/issues/207.

dustContributor:
Glad to help :smile: Could you please update the SNAPSHOT version? Most recent one is from last week.

decebals:
> Could you please update the SNAPSHOT version? 

Done.

PR #239, opened 2018-09-05
Update PropertiesPluginDescriptorFinder.java

heroyin:
Skip jar plugins
to fix   [org.pf4j.CompoundPluginDescriptorFinder.find(CompoundPluginDescriptorFinder.java:71)]  error log.


decebals:
@heroyin 
Can you reevaluate the problem in the context of new released PF4J 3? I did some improvements that I think that resolve your problem.

decebals:
This PR is relatively old and it's a big chance that the problem to be fixed until now. Please reopen it if you think that is still valuable.

PR #255, opened 2018-12-02
Adds ability to set plugin config directory

michaelruocco:
I have a use case where I am building a service with a number of plugins, this service is being deployed into docker. My original plan was to use the plugins directory as a volume mount, this way I can use a docker container and just mount the plugins I want to load into the container. However, I am using openshift which only allows a secret size of 1MB which is too small for some of my plugins, so instead I have had to take a different approach.

What I am doing now is building all the plugins directly into the docker image and using a volume to pick up an enabled.txt file to say which plugins should be enabled. This is fine except that the enabled.txt file needs to be in the same directory as the plugins themselves. If I want to use that directory as a volume (so that the config can be picked up dynamically) it means all the plugin jars are removed.

I am currently getting around this problem by mounting the enabled.txt file into a different folder location and then using a script to copy it into the plugins directory at startup. This works but is a bit fiddly, it would be easier if I could provide a system parameter to tell pf4j where to look for the enabled.txt file. This PR adds the ability to do this, I have just done it the quickest and easiest way I could think of, whilst also providing unit tests and it works as expected, but I am happy to refactor the changes if they don't fit with existing standards etc.

### DIFF
--- a/pf4j/pom.xml
+++ b/pf4j/pom.xml
@@ -55,6 +55,12 @@
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.19.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -620,6 +620,21 @@ public abstract class AbstractPluginManager implements PluginManager {
         return pluginsRoot;
     }
 
+    /**
+     * Add the possibility to override the plugins configuration root.
+     * If a {@code pf4j.pluginsConfigDir} system property is defined than this method returns that root.
+     * Otherwise returns the plugins root
+     *
+     * @return the plugins configuration root
+     */
+    public Path getPluginsConfigRoot() {
+        String configDir = System.getProperty("pf4j.pluginsConfigDir");
+        if (configDir == null) {
+            return pluginsRoot;
+        }
+        return Paths.get(configDir);
+    }
+
     @Override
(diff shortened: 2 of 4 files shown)


decebals:
@michaelruocco 
Thanks for your contribution.
Because your use case is not so common, I prefer to preserve the actual code.
If someone wants to modify the actual behavior of `DefaultPluginMnaager`, so to override the factory methods (`createXYZ`) is the correct approach.
In this use case the solution is:
```java
PluginManager pluginManager = new DefaultPluginManager() {

    @Override
    protected PluginStatusProvider createPluginStatusProvider() {
        Path configPath = <your business here>; 
        return new DefaultPluginStatusProvider(configPath);
    }

}
```

The idea is that the new `getPluginsConfigRoot` method in API, adds a new concept, `pluginsConfigRoot`, that makes sense only in the context of `DefaultPluginStatusProvider`. If I use other `PluginStatusProvider` like a database implementation, the `getPluginsConfigRoot` is a non sense.

michaelruocco:
Hmm okay, I hadn't actually thought of that approach. I will give it a try and let you know how I get on. Thanks for the suggestion.

If I can achieve the same result from extending the plugin manager, I may still want to contribute it back to either this project, or the pf4j-spring project. The main reason for this will be that I will be potentially be looking at using this in multiple projects, so would prefer not have duplicate this code in each one.

decebals:
If you insist to have this feature, please create a new PR that will contain your code with `System.getProperty` injected in `DefaultPluginManager.createPluginStatusProvider()`. No new method (private or public), only something minimal like:
```java
@Override
protected PluginStatusProvider createPluginStatusProvider() {
    String configDir = System.getProperty("pf4j.pluginsConfigDir");
    Path confiPath = configDir != null) ? Paths.get(configDir) : pluginRoot

    return new DefaultPluginStatusProvider(configPath);
}
```

michaelruocco:
@decebals thanks, that is pretty much what I have done, but I have added it on pf4j-spring instead, https://github.com/pf4j/pf4j-spring/pull/24.

PR #256, opened 2018-12-02
Adds ability to configure plugin directory

michaelruocco:
Have reimplemented following your simpler suggested method.


decebals:
Thanks!

PR #262, opened 2018-12-21
load extensions from plugin libraries

pinhead84:
The extension finder does not take plugin libraries (in the plugin's `lib` directory) into consideration. 

This pull request updates the `LegacyExtensionFinder` / `ServiceProviderExtensionFinder` class in order to scan the plugin libraries for `META-INF/extensions.idx` / `META-INF/services`.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
@@ -83,14 +83,21 @@ public class LegacyExtensionFinder extends AbstractExtensionFinder {
             Set<String> bucket = new HashSet<>();
 
             try {
-                URL url = ((PluginClassLoader) plugin.getPluginClassLoader()).findResource(getExtensionsResource());
-                if (url != null) {
-                    log.debug("Read '{}'", url.getFile());
-                    try (Reader reader = new InputStreamReader(url.openStream(), StandardCharsets.UTF_8)) {
-                        LegacyExtensionStorage.read(reader, bucket);
-                    }
-                } else {
+                Enumeration<URL> urls = ((PluginClassLoader) plugin.getPluginClassLoader()).findResources(getExtensionsResource());
+                if (urls == null || !urls.hasMoreElements()) {
                     log.debug("Cannot find '{}'", getExtensionsResource());
+                } else {
+                    while (urls.hasMoreElements()) {
+                        URL url = urls.nextElement();
+                        if (url != null) {
+                            log.debug("Read '{}'", url.getFile());
+                            try (Reader reader = new InputStreamReader(url.openStream(), StandardCharsets.UTF_8)) {
+                                LegacyExtensionStorage.read(reader, bucket);
+                            }
+                        } else {
+                            log.debug("Cannot find '{}'", getExtensionsResource());
+                        }
+                    }
                 }
 
                 debugExtensions(bucket);
--- a/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
@@ -89,13 +89,9 @@ public class LegacyExtensionFinder extends AbstractExtensionFinder {
                 } else {
                     while (urls.hasMoreElements()) {
                         URL url = urls.nextElement();
-                        if (url != null) {
-                            log.debug("Read '{}'", url.getFile());
-                            try (Reader reader = new InputStreamReader(url.openStream(), StandardCharsets.UTF_8)) {
-                                LegacyExtensionStorage.read(reader, bucket);
-                            }
-                        } else {
-                            log.debug("Cannot find '{}'", getExtensionsResource());
+                        log.debug("Read '{}'", url.getFile());
+                        try (Reader reader = new InputStreamReader(url.openStream(), StandardCharsets.UTF_8)) {
+                            LegacyExtensionStorage.read(reader, bucket);
                         }
                     }
                 }
--- a/pf4j/src/main/java/org/pf4j/ServiceProviderExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/ServiceProviderExtensionFinder.java
@@ -33,6 +33,7 @@ import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collections;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


decebals:
@pinhead84 
Firs of all thanks for you contribution.
Before analyse and merge this PR, can you explain us in few words what problem this PR resolves? Detailing your concrete use case is very welcome.

pinhead84:
@decebals 
At first I'd like to thank you for the great work and effort you've put into pf4j. I'm currently migrating my application from [JPF](http://jpf.sourceforge.net/) to pf4j - and it works pretty flawlessly. Your solution is well designed and I like its "keep it simple" approach very much.

Due to some compile time dependencies between my plugins it's necessary to put some extensions into a separate library, that is bundled with a plugin (in the plugins `lib` folder). I can give you more details about the dependency problems leading me to this approach, if you like to know. 

To make a long story short, JPF was able to find the extensions, that are bundled in the plugins `lib` folder - but pf4j currently just ignores them. Therefore I've implemented my own extension finder. It works as expected within my application and I could keep it private and maintain it for myself. But I think this fix might be helpful for others, too. Therefore I've decided to send you this pull request.

Please let me know, what you think about the pull request or if you have further questions about my proposal.

pinhead84:
This example describes my concrete use case for extensions in plugin libraries:

pf4j-optional-extensions-circular

It's the same scenario as described with issue #266. There are two plugins, that depend on each other. But because of the circular dependency it is **not possible** to compile the plugins with Java. Therefore I've moved both extensions into a separate package `com.example.plugins.contacts.addons`:

pf4j-optional-extensions

The `com.example.plugins.contacts.addons` package is compiled into a separate JAR library, which is added to the `lib` folder of the contacts plugin.

But currently pf4j does not look for extensions in the `lib` folder of any plugins. Therefore the `ContactsCalendarFormExtension` / `CalendarContactsFormExtension` provided by the library is ignored. The proposed pull request should fix that issue.

By the way: I think, that this is not the only use case / situation, where it might be helpful to provide extensions through a plugin library. Even if you're not interested in optional plugin dependencies this fix might be useful for others.

Please let me know, what you think about it.

decebals:
@pinhead84 
Thanks for detailed explanation. Now I understand very well your problem. At first sight I think that something it's wrong in your application, from design point of view, with that circular dependency but I cannot made a recommendation. Maybe an idea is to make `addons` package a plugin that has as dependencies  the other two. It's interesting to know how other plugin developers (Eclipse, Idea Intellij, ...) resolve this problem in an elegant mode.

I want to verify that your PR doesn't come with a performance penalty because it's a (very) particular case and I don't to affect all people.

pinhead84:
> At first sight I think that something it's wrong in your application, from design point of view, with that circular dependency but I cannot made a recommendation. Maybe an idea is to make addons package a plugin that has as dependencies the other two. 

I also thought, that my approach might be the root of the problem. But a separate plugin, that provides the addons package is not really a solution in my case. I've described the problem about your proposed workaround in issue #266:

> This might be an acceptable solution for small application. But if we think a little bit bigger, this approach might lead to a huge number of plugins only to reflect those inner plugin dependencies.
> 
> A number of **n** application plugins would require **n * (n-1) / 2** additional plugins to reflect all possible inner plugin dependencies.

My application currently provides 7 plugins with certain inner dependencies. With pf4j in its current state I would have to create round about **10 more plugins** only to match those inner plugin dependencies. This number might increase even more for every new plugin, that is added to the application. I don't think, that this would be a better solution for my particular case. ;)

---

> It's interesting to know how other plugin developers (Eclipse, Idea Intellij, ...) resolve this problem in an elegant mode.

I don't know how Eclipse or IntelliJ are solving this issue. Maybe they don't need these kind of features. But JPF provides optional plugin dependencies as described in issue #266, which solve these kind of problems quite well.

But to make these optional dependencies work in my case, I would also need this pull request. Otherwise I wouldn't been able to load the extensions from the addons library (in the plugins `lib` directory).

---

> I want to verify that your PR doesn't come with a performance penalty because it's a (very) particular case and I don't to affect all people.

I fully agree - performance testing is always a good idea and the fix should not have a negative impact on other users. I haven't done any performance tests, yet. But on the other hand I didn't see a big increase in startup time of my application after the fix was implemented.

decebals:
> But JPF provides optional plugin dependencies as described in issue #266, which solve these kind of problems quite well.
> 
> But to make these optional dependencies work in my case, I would also need this pull request. Otherwise I wouldn't been able to load the extensions from the addons library (in the plugins lib directory).

It's not clear form me the relation between this PR and #266. In your first paragraph from quote you says that #266 solves your problem. On the other hand, in the next paragraph you say that you need this PR to solve your problem.

pinhead84:
Sorry, I didn't want to confuse you. Basically there are two problems:

- Issue #266 solves the problem, that plugins might have an optional dependency on each other. The contacts plugin should be available in my application even if the calendar plugin was not installed / enabled.

- Another problem is solved by this pull request. To avoid circular dependencies on compilation I have to move the `com.example.plugins.contacts.addons` package into a separate Java library, which is placed in the `lib` folder of the contacts plugin. PF4J must be able to read the extensions provided by this library (which is currently not possible).

Both problems are more or less independent from each other. But in order to realize the described design pattern with PF4J, I need both of them to be fixed.

pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
```diff
@@ -83,14 +83,21 @@ public LegacyExtensionFinder(PluginManager pluginManager) {
...
             try {
-                URL url = ((PluginClassLoader) plugin.getPluginClassLoader()).findResource(getExtensionsResource());
-                if (url != null) {
-                    log.debug("Read '{}'", url.getFile());
-                    try (Reader reader = new InputStreamReader(url.openStream(), StandardCharsets.UTF_8)) {
-                        LegacyExtensionStorage.read(reader, bucket);
-                    }
-                } else {
+                Enumeration<URL> urls = ((PluginClassLoader) plugin.getPluginClassLoader()).findResources(getExtensionsResource());
+                if (urls == null || !urls.hasMoreElements()) {
```
decebals: I think that `urls` cannot be null. The javadoc is not clear so we can keep null check.

pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
```diff
@@ -83,14 +83,21 @@ public LegacyExtensionFinder(PluginManager pluginManager) {
...
                     log.debug("Cannot find '{}'", getExtensionsResource());
+                } else {
+                    while (urls.hasMoreElements()) {
+                        URL url = urls.nextElement();
+                        if (url != null) {
```
decebals: Url is not null, so no need for null check

pf4j/src/main/java/org/pf4j/ServiceProviderExtensionFinder.java
```diff
@@ -94,18 +95,25 @@ public ServiceProviderExtensionFinder(PluginManager pluginManager) {
...
-
-                    bucket.addAll(readExtensions(extensionPath));
-                } else {
+                Enumeration<URL> urls = ((PluginClassLoader) plugin.getPluginClassLoader()).findResources(getExtensionsResource());
+                if (urls == null || !urls.hasMoreElements()) {
                     log.debug("Cannot find '{}'", getExtensionsResource());
+                } else {
+                    while (urls.hasMoreElements()) {
+                        URL url = urls.nextElement();
+                        if (url != null) {
```
decebals: The same thing (no need for null check)

pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
pinhead84: You're right. This null check is definitely not necessary. I've [updated the pull request](https://github.com/pf4j/pf4j/pull/262/commits/23a4263eef7de59ed4a5bb29f84af4010b749d75) accordingly.

pf4j/src/main/java/org/pf4j/ServiceProviderExtensionFinder.java
pinhead84: You're right. This null check is definitely not necessary. I've [updated the pull request](https://github.com/pf4j/pf4j/pull/262/commits/23a4263eef7de59ed4a5bb29f84af4010b749d75) accordingly.

decebals:
Thanks!

Issue #264, opened 2018-12-27
annotation processor creates wrong files under certain conditions


pinhead84:
I'm currently migrating my application from [JPF](http://jpf.sourceforge.net/) to pf4j and discovered a problem in the `ExtensionAnnotationProcessor` in combination with the `ServiceProviderExtensionStorage`.

I'm using the following pattern to define extension points and extensions:

![pf4j-extensions](https://user-images.githubusercontent.com/7954406/50482680-381d0780-09e8-11e9-8b47-f6d2a5551a5c.png)

In contrast to your examples there are abstract classes (`MainMenuAdapter`, `BootAdapter`) for each available extension point. The extensions are derived from these abstract classes instead of the extension point interfaces (`MainMenuExtensionPoint`, `BootExtensionPoint`). This gives a bit more flexibility, because later added methods in an extension point can be implemented in the abstract classes without the need of modifying the extension points in all plugins. From my experience in certain projects this is a quite common pattern - similar to the `java.awt.event.WindowListener` / `java.awt.event.WindowAdapter` approach in the Java classpath.

Basically this approach also works with pf4j. But the `ExtensionAnnotationProcessor` in combination with the `ServiceProviderExtensionStorage` doesn't behave as expected in this scenario. For the above scenario the following files are generated:

```
META-INF/services/com.example.api.BootAdapter
META-INF/services/com.example.api.MainMenuAdapter
```

Instead the `ExtensionAnnotationProcessor` should create these files:

```
META-INF/services/com.example.api.BootExtensionPoint
META-INF/services/com.example.api.MainMenuExtensionPoint
```

Because of this problem the `PluginManager` is not able to load the extensions by its interface via `pluginManager.getExtensions(BootExtensionPoint.class)`. Instead I would have to call `pluginManager.getExtensions(BootAdapter.class)`.

I'm going to propose a pull request with a possible workaround for this problem.

decebals:
Thanks again for detailed explanation. Now, the case with adapters doesn't work. The adapters are see as extensions by PF4J.
Your approach proposed in #265 is good. The same approach I use in [Pippo](https://github.com/pippo-java/pippo/blob/b446c6debc8d9136c2b3e9ff4a8ee247dcdaf50f/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowServer.java#L63) to declare extensions.
So, after some reviews and comments, #265 will be merged and the issue will be resolved.

PR #265, opened 2018-12-27
explicitly configure extension points for an extension

pinhead84:
This pull requests provides a possible workaround for issue #264.

An extension may provide its implemented extension points explicitly through its annotation. According to my example in issue #264 you might use:

```java
@Extension(provides = {BootExtensionPoint.class})
public class Plugin1BootExtension extends BootAdapter {
    // ... some implementations 
}
```

If the `@Extension` provides its implemented extension points through the `provides` attribute, there is no need for the annotation processor to detect the extension points automatically, which currently fails under the described conditions.

I know, that this might not the best approach. I guess you also can improve the automatic detection of extension points in the [`findExtensionPoints`](https://github.com/pf4j/pf4j/blob/ab12f7fa5a95dd570f2e62812a243d48411c1dce/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java#L168-L191) method. But with this pull request I don't want to break compatibility by changing your current implementation. Therefore I've decided not to change your implementation and keep it as a default behaviour.

In general I'm not sure, if an improved [`findExtensionPoints`](https://github.com/pf4j/pf4j/blob/ab12f7fa5a95dd570f2e62812a243d48411c1dce/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java#L168-L191) method would work in every possible use case as expected. Therefore it might be quite helpful to have a solution, that circumvents the automatic detection.

Please let me know, what you think about this solution.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/Extension.java
+++ b/pf4j/src/main/java/org/pf4j/Extension.java
@@ -34,4 +34,13 @@ public @interface Extension {
 
     int ordinal() default 0;
 
+    /**
+     * An array of extension points, that are implemented by this extension.
+     * This explicit configuration overrides the automatic detection of extension points in the
+     * {@link org.pf4j.processor.ExtensionAnnotationProcessor}.
+     *
+     * @return classes of extension points implemented by this extension
+     */
+    Class<? extends ExtensionPoint>[] provides() default {};
+
 }
--- a/pf4j/src/main/java/org/pf4j/Extension.java
+++ b/pf4j/src/main/java/org/pf4j/Extension.java
@@ -38,8 +38,12 @@ public @interface Extension {
      * An array of extension points, that are implemented by this extension.
      * This explicit configuration overrides the automatic detection of extension points in the
      * {@link org.pf4j.processor.ExtensionAnnotationProcessor}.
+     * <p>
+     * In case your extension is directly derived from an extension point this attribute is NOT required.
+     * But under certain <a href="https://github.com/pf4j/pf4j/issues/264">more complex scenarios</a> it
+     * might be useful to explicitly set the extension points for an extension.
      *
-     * @return classes of extension points implemented by this extension
+     * @return classes of extension points, that are implemented by this extension
      */
     Class<? extends ExtensionPoint>[] points() default {};
 
--- a/pf4j/src/main/java/org/pf4j/Extension.java
+++ b/pf4j/src/main/java/org/pf4j/Extension.java
@@ -41,6 +41,6 @@ public @interface Extension {
      *
      * @return classes of extension points implemented by this extension
      */
-    Class<? extends ExtensionPoint>[] provides() default {};
+    Class<? extends ExtensionPoint>[] points() default {};
 
 }
--- a/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
+++ b/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
@@ -17,6 +17,7 @@ package org.pf4j.processor;
 
 import org.pf4j.Extension;
 import org.pf4j.ExtensionPoint;
+import org.pf4j.util.ClassUtils;
 
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.ProcessingEnvironment;
@@ -25,7 +26,6 @@ import javax.lang.model.SourceVersion;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
-import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
@@ -104,9 +104,9 @@ public class ExtensionAnnotationProcessor extends AbstractProcessor {
             List<TypeElement> extensionPointElements;
 
             // use extension points, that were explicitly set in the extension annotation
-            AnnotationMirror annotation = getAnnotationMirror(extensionElement, Extension.class);
+            AnnotationMirror annotation = ClassUtils.getAnnotationMirror(extensionElement, Extension.class);
             AnnotationValue annotatedExtensionPoints = (annotation != null) ?
-                getAnnotationValue(annotation, "provides") :
+                ClassUtils.getAnnotationValue(annotation, "provides") :
                 null;
             List<? extends AnnotationValue> extensionPointClasses = (annotatedExtensionPoints != null) ?
                 (List<? extends AnnotationValue>) annotatedExtensionPoints.getValue() :
@@ -252,28 +252,5 @@ public class ExtensionAnnotationProcessor extends AbstractProcessor {
         }
 
         return storage;
-    }
-
-    private static AnnotationMirror getAnnotationMirror(TypeElement typeElement, Class<?> clazz) {
-        // get annotation of a type element
-        // as described at https://stackoverflow.com/a/10167558
-        String clazzName = clazz.getName();
-        for (AnnotationMirror m : typeElement.getAnnotationMirrors()) {
-            if (m.getAnnotationType().toString().equals(clazzName)) {
-                return m;
-            }
-        }
-        return null;
-    }
-
-    private static AnnotationValue getAnnotationValue(AnnotationMirror annotationMirror, String key) {
-        // get annotation value of a type element
-        // as described at https://stackoverflow.com/a/10167558
-        for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry : annotationMirror.getElementValues().entrySet()) {
-            if (entry.getKey().getSimpleName().toString().equals(key)) {
-                return entry.getValue();
-            }
-        }
-        return null;
     }
 }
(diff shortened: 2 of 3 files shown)


pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
```diff
@@ -229,4 +254,26 @@ private ExtensionStorage createStorage() {
         return storage;
     }
 
+    private static AnnotationMirror getAnnotationMirror(TypeElement typeElement, Class<?> clazz) {
```
decebals: Move this method in https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/util/ClassUtils.java

pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
```diff
@@ -229,4 +254,26 @@ private ExtensionStorage createStorage() {
...
+        String clazzName = clazz.getName();
+        for (AnnotationMirror m : typeElement.getAnnotationMirrors()) {
+            if (m.getAnnotationType().toString().equals(clazzName)) {
+                return m;
+            }
+        }
+        return null;
+    }
+
+    private static AnnotationValue getAnnotationValue(AnnotationMirror annotationMirror, String key) {
```
decebals: Move this method in https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/util/ClassUtils.java

pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
```diff
@@ -98,7 +101,29 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
 
             TypeElement extensionElement = (TypeElement) element;
 //            Extension annotation = element.getAnnotation(Extension.class);
-            List<TypeElement> extensionPointElements = findExtensionPoints(extensionElement);
+            List<TypeElement> extensionPointElements;
```
decebals: Try to create a nice generic static method with parameter(s) in Move this method in https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/util/ClassUtils.java and use that method here

decebals:
I prefer, as plugin developers, the short version
```java
@Extension(BootExtensionPoint.class)
public class Plugin1BootExtension extends BootAdapter {
    // ... some implementations 
}
```

instead of
```java
@Extension(provides = {BootExtensionPoint.class})
public class Plugin1BootExtension extends BootAdapter {
    // ... some implementations 
}
```

Something like https://github.com/pippo-java/pippo/blob/b446c6debc8d9136c2b3e9ff4a8ee247dcdaf50f/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowServer.java#L63. The approach used in Pippo is proved to be good.

pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
pinhead84: I'm not sure, if I understand your change request correctly. Does PF4J or some plugin developer need the detection of extension points as global (final & static) method in `ClassUtils`?

I'm assuming, that only the `ExtensionAnnotationProcessor` would ever need this mechanism. It seems more logic to me to put this modification into the private `findExtensionPoints` method, where the automatic detection is already present. I've [updated my pull request](https://github.com/pf4j/pf4j/pull/265/commits/3d929a3d8831ba2acccd31182f9308b8763254d3) to illustrate, how I would approach this.

But if you prefer to have some (or all?) parts of the extension point detection as public method in `ClassUtils`, I will make the necessary changes of course.

pf4j/src/main/java/org/pf4j/Extension.java
```diff
@@ -34,4 +34,13 @@
 
     int ordinal() default 0;
 
+    /**
+     * An array of extension points, that are implemented by this extension.
+     * This explicit configuration overrides the automatic detection of extension points in the
+     * {@link org.pf4j.processor.ExtensionAnnotationProcessor}.
```
decebals: I think that here we must add some explanation lines when to use this attribute (your approach with Adapter is very good as example - also my example from [Pippo](https://github.com/pippo-java/pippo/blob/b446c6debc8d9136c2b3e9ff4a8ee247dcdaf50f/pippo-server-parent/pippo-undertow/src/main/java/ro/pippo/undertow/UndertowServer.java#L63) is very clear as use case).

pf4j/src/main/java/org/pf4j/Extension.java
```diff
@@ -34,4 +34,13 @@
...
+     *
+     * @return classes of extension points implemented by this extension
+     */
+    Class<? extends ExtensionPoint>[] provides() default {};
```
decebals: Here I have two questions:
- do you think that it's better to rename this attribute from `provides` in `points` (I feel that it's better from semantically point of view) ?
- I don't have until now an extension that implements multiple points. I don't know if it's possible with the actual release and I didn't tested. Personal I prefer in my projects to have one extension for one extension point. So, my question is do we real need a list of extension point class for this attribute, is not a single extension point class enough (at least for the moment)?

pf4j/src/main/java/org/pf4j/util/ClassUtils.java
```diff
@@ -72,6 +77,60 @@ public static List getAllAbstractClasses(Class aClass) {
...
     */
 
+    /**
+     * Get a certain annotation of a {@link TypeElement}.
+     * See <a href="https://stackoverflow.com/a/10167558">stackoverflow.com</a> for more information.
+     *
+     * @param typeElement the type element, that contains the requested annotation
+     * @param annotationClass the class of the requested annotation
+     * @return the requested annotation or null, if no annotation of the provided class was found
+     * @throws NullPointerException if <code>typeElement</code> or <code>annotationClass</code> is null
```
decebals: Do we need this line (throws) in javadoc ? It's uncommon for me to add a NPE in javadoc's `throws` section. It's clear for me that the parameters cannot be null. In the end we can add this clarification in @param.

pf4j/src/main/java/org/pf4j/Extension.java
pinhead84: I've made some changes to the documentation and [updated my pull request](https://github.com/pf4j/pf4j/pull/265/commits/0de925f7532c470e39c5460626f88de4b9cf3a8d). From my point of view it would be better to document this attribute in the [official documentation](https://pf4j.org/). The Javadoc could contain a link, that points to the corresponding page of the documentation website.

It's much easier to document those kind of issues together with images in the official documentation compared to a somewhat limited javadoc. Also I guess most users would expect to find those kind of information in the official documentation instead of the Javadocs.

Because the new attribute is currently not documented yet, I've just added a link to issue #264 to describe the use case. Maybe you can replace that link later or you can add another link to Pippo.

pf4j/src/main/java/org/pf4j/Extension.java
pinhead84: > do you think that it's better to rename this attribute from `provides` in `points` (I feel that it's better from semantically point of view) ?

I'm also fine with `points`. I guess that's even more precise. I've [updated the pull request](https://github.com/pf4j/pf4j/pull/265/commits/33558c798efa0238ebdd30dab018be5570654c8f) accordingly.

> I don't have until now an extension that implements multiple points. Personal I prefer in my projects to have one extension for one extension point. 

It's the same for me. I basically always use the pattern described in issue #264 and don't make use of multiple extension points in a single extension class.

> So, my question is do we real need a list of extension point class for this attribute, is not a single extension point class enough (at least for the moment)?

Even if I don't use this approach, I would suggest to keep the possibility to use multiple extension points in a single extension. 

Java allows multiple inheritance of interfaces. As far as I know this feature used from time to time. If PF4J wouldn't support this approach with its extension, it would restrict plugin developers in their possibilities. As long as there is no definitive technical reason of not supporting this feature, I would suggest to support it with PF4J.

> I don't know if it's possible with the actual release and I didn't tested.

I also didn't test it. But as far as I've seen / understood the source code of PF4J, I guess it should / could work. Why don't we just add this scenario to the demo application? - Afterwards we should no better. ;)

pf4j/src/main/java/org/pf4j/util/ClassUtils.java
pinhead84: In general I think its useful to know about runtime exceptions thrown by a method without looking into the source code. If the plugin developer uses Eclipse / Netbeans / IntelliJ he should see the javadoc commment directly in his IDE. He wouldn't have to make assumptions about possible NPE's or had to look into the PF4J source code.

Of course you might also add this information into the `@param`. But I personally prefer to separate those kind of information. Theoretically there might be other runtime exceptions besides a NPE, too (e.g. `IllegalArgumentException`).

By the way: The Java API documentation also contains the runtime exception thrown by a method (e.g. [`Integer#valueOf(java.lang.String)`](https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#valueOf(java.lang.String))). Therefore I assume, that this approach is quite common.

But I leave this up to you. If you prefer another way to write Javadocs, feel free to change them after you've merged this pull request or let me know about your preferred conventions.

decebals:
Thanks!

Issue #266, opened 2018-12-27
support optional plugin dependencies


pinhead84:
Under certain situations it might be very helpful to support optional dependencies between plugins. For example the following scenario is currently pretty difficult to realize with pf4j:

![pf4j-optional-extensions](https://user-images.githubusercontent.com/7954406/50489511-5266dd80-0a08-11e9-8e3b-31c4afa71115.png)

One plugin adds contact management and the other plugin adds calendar functionality to an application. Both plugins provide a form to modify contacts / calendar entries, that can be extended by other plugins through the `ContactsFormExtensionPoint` / `CalendarFormExtensionPoint`.

Let's assume, that the contacts plugin adds a panel to the calendar form, that shows assigned contacts for a certain calendar entry. On the other hand the calendar plugin adds a panel to the contacts form, that shows assigned calendar entries for a certain contact.

In this case, both plugins would depend on each other. I would have to define these dependencies in their `META-INF/MANIFEST.MF` accordingly. 

But what happens, if one of the plugins is missing, because the application user might not have installed it? - Currently pf4j wouldn't load the contacts plugin, if the calendar plugin is missing and vice versa (due to the defined dependencies in `META-INF/MANIFEST.MF`). But in this case, the contacts plugin should still be usable - even if the calendar plugin is not present.

---

With the current version of pf4j I only see one solution for this problem. I would have to remove the dependencies between the plugins and have to create a **separate plugin**, that provides the `com.example.plugins.contacts.addons` package. This additional plugin would only be loaded, if contacts **and** calendar plugin are present.

This might be an acceptable solution for small application. But if we think a little bit bigger, this approach might lead to a huge number of plugins only to reflect those inner plugin dependencies. 

- Four application plugins would require six additional plugins to reflect all possible inner plugin dependencies.
- Six application plugins  would require fifteen additional plugins to reflect all possible inner plugin dependencies.
- Or in general: A number of **n** application plugins would require **n * (n-1) / 2** additional plugins to reflect all possible inner plugin dependencies.

I hope my math is right. ;) - But even if the formula is wrong, it should show, that this possible workaround can lead to a big mess on more complex applications with a bigger number of plugins.

---

But how did others solve this problem? - For example [JPF](http://jpf.sourceforge.net/) solved this problem by allowing **optional plugin dependencies**. I think, this would also be the most elegant solution for pf4j.

[...]

PR #270, opened 2019-01-03
Optional plugin dependencies

pinhead84:
This pull request proposes a solution for issue #266.

### How optional plugin dependencies are defined

According to my example in issue #266 the calendar plugin has an optional dependency to the contacts plugin. I suggest to add a question mark after the plugin name to define those optional dependencies:

```xml
<plugin.dependencies>Contacts?</plugin.dependencies>
```

The manifest of the calendar plugin looks like this:

```
Plugin-Class: com.example.plugins.calendar.CalendarPlugin
Plugin-Dependencies: Contacts?
Plugin-Id: Calendar
Plugin-License: Example license
Plugin-Provider: Example company
Plugin-Version: 1.0.0
```

Vice versa the contacts plugin has an optional dependency to the calendar plugin:

```
Plugin-Class: com.example.plugins.contacts.ContactsPlugin
Plugin-Dependencies: Calendar?
Plugin-Id: Contacts
Plugin-License: Example license
Plugin-Provider: Example company
Plugin-Version: 1.0.0
```

If a plugin depends on a particular version of another plugin you can write:

```xml
<plugin.dependencies>Contacts?@1.0.0</plugin.dependencies>
```

This naming scheme is implemented with [this commit](https://github.com/pf4j/pf4j/pull/270/commits/2f549f9164437071c60dd8a48d52257dd3e1fb85). Please let me know, if you prefer other conventions here. It doesn't really matter to me, how to define this exactly, as long as we can define it somehow. ;)

By the way: I'm currently not sure, if an optional plugin dependency with a specific version number is processed as expected with my pull request. This special case might need some more testing on my side.

### How optional plugin dependencies are processed

The `DependencyResolver` doesn't put optional plugin dependencies into the dependencies graph. This avoids possible problems with circular dependencies and also makes sure, that a plugin is loaded even if one of its optional dependencies is not resolved. This is implemented with [this commit](https://github.com/pf4j/pf4j/pull/270/commits/0ffcc350e88300359fc306e269377c607c9e3593). Maybe there is a more elegant solution...

In case an optional plugin dependency is not available at runtime, the `PluginClassLoader` should not process its class loader. Otherwise a `NullPointerException` might occur. This is implemented with [this commit](https://github.com/pf4j/pf4j/pull/270/commits/9fd2016be4cf8534ff81d13bca65bf2d0d6dafe1).

### How extensions are processed for missing optional plugins

According to the example in issue #266 there is a separate Java library, that provides the `com.example.plugins.contacts.addons` package and is bundled in the `lib` directory of the contacts plugin. 

In case that the calendar plugin is not installed / activated by the user, PF4J would throw an `NoClassDefFoundError` during startup. PF4J is not able to load the `com.example.plugins.contacts.addons.CalendarContactsFormExtension` class through the class loader, because the required calendar plugin (and its provided classes) can't be resolved at runtime.

Ignoring those `NoClassDefFoundError`'s might be possible, but is a quite ugly solution. I guess it's not a good coding style, to simply ignore those kind of errors.

I see only one clean solution for that problem, that has been addressed similar by [JPF](http://jpf.sourceforge.net/). PF4J needs to know **before** an extension is loaded through the class loader, on which plugins the extension depends. If the dependent plugins are not available / enabled at runtime, the corresponding extension should not be loaded.

For this particular use case I've added another parameter to the `@Extension` annotation. The extension points in my example are defined like this:

```java
@Extension(plugins = {CalendarPlugin.ID, ContactsPlugin.ID})
public class CalendarContactsFormExtension implements ContactsFormExtensionPoint {
   // ... some implementations ...
}
```

```java
@Extension(plugins = {CalendarPlugin.ID, ContactsPlugin.ID})
public class ContactsCalendarFormExtension implements CalendarFormExtensionPoint {
   // ... some implementations ...
}
```

The additional annotation parameter is integrated with [this commit](https://github.com/pf4j/pf4j/pull/270/commits/e5f217227b470163d21763fc920bb96f63e04d89).

Before the `AbstractExtensionFinder` loads the extension through the class loader, it loads the `@Extension` annotation and checks the configured `plugins` against the current plugin environment. Only if all plugins defined by an extension are available / started at runtime, the corresponding extension class is loaded by `AbstractExtensionFinder`. This mechanism is implemented with [this commit](https://github.com/pf4j/pf4j/pull/270/commits/891179e63b993ae5aa2d8d8e23279c4ff64a2a64).

By the way: This approach is backwards compatible. If plugin developers don't use the `plugins` parameter in their `@Extension`'s, PF4J will behave exactly like before.

### PF4J requires ASM for bytecode analysis

Unfortunately I have not found a way, to load the `@Extension` annotation for classes with missing dependencies. In case that an optional plugin dependency is missing, a `NoClassDefFoundError` is thrown, if the class is loaded by `AbstractExtensionFinder`. But without having the extension class loaded I also can't get the `@Extension` annotation for the required verification (as described above) via [`Class#getAnnotation(java.lang.Class)`](https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getAnnotation(java.lang.Class)).

Please let me know, if you know a solution to load an `@Extension` annotation for a class with missing dependencies. 

JPF for example does not have this problem, because they're defining those dependencies in separate XML files, which of course can easily be loaded before loading the extension class itself. But PF4J uses annotations for those specifications, which I highly prefer but which also makes it more difficult to access those information without loading the class itself.

From my point of view, a simple solution for this problem might be a bytecode analysis with [ASM](https://asm.ow2.io). The `@Extension` annotation can be parsed by ASM without loading the extension class into the class loader (which might lead to `NoClassDefFoundError`'s). This mechanism is implemented with [this commit](https://github.com/pf4j/pf4j/pull/270/commits/891179e63b993ae5aa2d8d8e23279c4ff64a2a64).

### Notes about the ASM dependency

I've configured the asm depdency as optional in the `pom.xml`. PF4J will still work for most users without this dependency. It is only necessary, if 

- optional plugin dependencies are used. In this case `AbstractExtensionFinder` will automatically enable this feature.
- the plugin developer uses the `plugins` parameter in the `@Extension` annotation. In this case he might enable this feature manually by calling `AbstractExtensionFinder#setCheckForExtensionDependencies(boolean checkForExtensionDependencies)`.

In any other cases, the ASM dependency is not required in order to use PF4J. 

I know, that it is not optimal to add another dependency to PF4j. But the `asm-7.0.jar` file is pretty small (113 KB), has an acceptable 3 clause BSD license and is not required for most of the users. Therefore I hope it's acceptable for you to add this optional dependency to PF4J.

---

I'm currently testing this implementation within my application. It currently behaves as expected and does not seem to have a relevant impact on the performance.

Please let me know, what you think about this approach. Any suggestions or improvements are highly appreciated. :)

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
@@ -15,12 +15,14 @@
  */
 package org.pf4j;
 
+import org.pf4j.asm.ExtensionInfo;
 import org.pf4j.util.ClassUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -35,9 +37,29 @@ public abstract class AbstractExtensionFinder implements ExtensionFinder, Plugin
 
     protected PluginManager pluginManager;
     protected volatile Map<String, Set<String>> entries; // cache by pluginId
+    protected volatile Map<String, ExtensionInfo> extensionInfos; // cache extension infos by class name
+    protected boolean checkForExtensionDependencies = false;
 
     public AbstractExtensionFinder(PluginManager pluginManager) {
         this.pluginManager = pluginManager;
+
+        // By default we're assuming, that no checks for extension dependenciess are necessary.
+        //
+        // But if one of the available plugins defines an optional plugin dependency,
+        // extensions might not be loadable due to missing optional dependencies.
+        // Therefore we're enabling a check of optional extensions, if at least one
+        // optional plugin dependency was specified.
+        for (PluginWrapper plugin : this.pluginManager.getPlugins()) {
+            for (PluginDependency dependency : plugin.getDescriptor().getDependencies()) {
+                if (dependency.isOptional()) {
+                    checkForExtensionDependencies = true;
+                    break;
+                }
+            }
+            if (checkForExtensionDependencies) {
+                break;
+            }
+        }
     }
 
     public abstract Map<String, Set<String>> readPluginsStorages();
@@ -97,6 +119,41 @@ public abstract class AbstractExtensionFinder implements ExtensionFinder, Plugin
 
         for (String className : classNames) {
             try {
+                if (isCheckForExtensionDependencies()) {
+                    // Load extension annotation without initializing the class itself.
+                    //
+                    // If optional dependencies are used, the class loader might not be able
+                    // to load the extension class because of missing optional dependencies.
+                    //
+                    // Therefore we're extracting the extension annotation via asm, in order
+                    // to extract the required plugins for an extension. Only if all required
+                    // plugins are currently available and started, the corresponding
+                    // extension is loaded through the class loader.
+                    ExtensionInfo extensionInfo = getExtensionInfo(className, classLoader);
+                    if (extensionInfo == null) {
+                        log.error("No extension annotation was found for '{}'", className);
+                        continue;
+                    }
+
+                    // Make sure, that all plugins required by this extension are available.
+                    List<String> missingPluginIds = new ArrayList<>();
+                    for (String requiredPluginId : extensionInfo.getPlugins()) {
+                        PluginWrapper requiredPlugin = pluginManager.getPlugin(requiredPluginId);
+                        if (requiredPlugin == null || !PluginState.STARTED.equals(requiredPlugin.getPluginState())) {
+                            missingPluginIds.add(requiredPluginId);
+                        }
+                    }
+                    if (!missingPluginIds.isEmpty()) {
+                        StringBuilder missing = new StringBuilder();
+                        for (String missingPluginId : missingPluginIds) {
+                            if (missing.length() > 0) missing.append(", ");
+                            missing.append(missingPluginId);
+                        }
+                        log.trace("Extension '{}' is ignored due to missing plugins: {}", className, missing);
+                        continue;
+                    }
+                }
+
                 log.debug("Loading class '{}' using class loader '{}'", className, classLoader);
                 Class<?> extensionClass = classLoader.loadClass(className);
 
@@ -188,6 +245,44 @@ public abstract class AbstractExtensionFinder implements ExtensionFinder, Plugin
         entries = null;
     }
 
+    /**
+     * Returns true, if the extension finder checks extensions for its required plugins.
+     * This feature has to be enabled, in order check the availability of
+     * {@link Extension#plugins()} configured by an extension.
+     * <p>
+     * This feature is enabled by default, if at least one available plugin makes use of
+     * optional plugin dependencies. Those optional plugins might not be available at runtime.
+     * Therefore any extension is checked by default against available plugins before its
+     * instantiation.
+     * <p>
+     * Notice: This feature requires the optional <a href="https://asm.ow2.io/">ASM library</a>
+     * to be available on the applications classpath.
+     *
+     * @return true, if the extension finder checks extensions for its required plugins
+     */
+    public final boolean isCheckForExtensionDependencies() {
+        return checkForExtensionDependencies;
+    }
+
+    /**
+     * Plugin developers may enable / disable checks for required plugins of an extension.
+     * This feature has to be enabled, in order check the availability of
+     * {@link Extension#plugins()} configured by an extension.
+     * <p>
+     * This feature is enabled by default, if at least one available plugin makes use of
+     * optional plugin dependencies. Those optional plugins might not be available at runtime.
+     * Therefore any extension is checked by default against available plugins before its
+     * instantiation.
+     * <p>
+     * Notice: This feature requires the optional <a href="https://asm.ow2.io/">ASM library</a>
+     * to be available on the applications classpath.
+     *
+     * @param checkForExtensionDependencies true to enable checks for optional extensions, otherwise false
+     */
+    public void setCheckForExtensionDependencies(boolean checkForExtensionDependencies) {
+        this.checkForExtensionDependencies = checkForExtensionDependencies;
+    }
+
     protected void debugExtensions(Set<String> extensions) {
         if (log.isDebugEnabled()) {
             if (extensions.isEmpty()) {
@@ -216,6 +311,34 @@ public abstract class AbstractExtensionFinder implements ExtensionFinder, Plugin
         }
 
         return entries;
+    }
+
+    /**
+     * Returns the parameters of an {@link Extension} annotation without loading
+     * the corresponding class into the class loader.
+     *
+     * @param className name of the class, that holds the requested {@link Extension} annotation
+     * @param classLoader class loader to access the class
+     * @return the contents of the {@link Extension} annotation or null, if the class does not
+     * have an {@link Extension} annotation
+     */
+    private ExtensionInfo getExtensionInfo(String className, ClassLoader classLoader) {
+        if (extensionInfos == null) {
+            extensionInfos = new HashMap<>();
+        }
+
+        if (!extensionInfos.containsKey(className)) {
+            log.trace("Load annotation for '{}' using asm", className);
+            ExtensionInfo info = ExtensionInfo.load(className, classLoader);
+            if (info == null) {
+                log.warn("No extension annotation was found for '{}'", className);
+                extensionInfos.put(className, null);
+            } else {
+                extensionInfos.put(className, info);
+            }
+        }
+
+        return extensionInfos.get(className);
     }
 
     private ExtensionWrapper createExtensionWrapper(Class<?> extensionClass) {
--- a/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
@@ -38,28 +38,10 @@ public abstract class AbstractExtensionFinder implements ExtensionFinder, Plugin
     protected PluginManager pluginManager;
     protected volatile Map<String, Set<String>> entries; // cache by pluginId
     protected volatile Map<String, ExtensionInfo> extensionInfos; // cache extension infos by class name
-    protected boolean checkForExtensionDependencies = false;
+    protected Boolean checkForExtensionDependencies = null;
 
     public AbstractExtensionFinder(PluginManager pluginManager) {
         this.pluginManager = pluginManager;
-
-        // By default we're assuming, that no checks for extension dependenciess are necessary.
-        //
-        // But if one of the available plugins defines an optional plugin dependency,
-        // extensions might not be loadable due to missing optional dependencies.
-        // Therefore we're enabling a check of optional extensions, if at least one
-        // optional plugin dependency was specified.
-        for (PluginWrapper plugin : this.pluginManager.getPlugins()) {
-            for (PluginDependency dependency : plugin.getDescriptor().getDependencies()) {
-                if (dependency.isOptional()) {
-                    checkForExtensionDependencies = true;
-                    break;
-                }
-            }
-            if (checkForExtensionDependencies) {
-                break;
-            }
-        }
     }
 
     public abstract Map<String, Set<String>> readPluginsStorages();
@@ -243,6 +225,22 @@ public abstract class AbstractExtensionFinder implements ExtensionFinder, Plugin
         // TODO optimize (do only for some transitions)
         // clear cache
         entries = null;
+
+        // By default we're assuming, that no checks for extension dependencies are necessary.
+        //
+        // A plugin, that has an optional dependency to other plugins, might lead to unloadable
+        // Java classes (NoClassDefFoundError) at application runtime due to possibly missing
+        // dependencies. Therefore we're enabling the check for optional extensions, if the
+        // started plugin contains at least one optional plugin dependency.
+        if (checkForExtensionDependencies == null && PluginState.STARTED.equals(event.getPluginState())) {
+            for (PluginDependency dependency : event.getPlugin().getDescriptor().getDependencies()) {
+                if (dependency.isOptional()) {
+                    log.debug("Enable check for extension dependencies via ASM.");
+                    checkForExtensionDependencies = true;
+                    break;
+                }
+            }
+        }
     }
 
     /**
@@ -261,7 +259,7 @@ public abstract class AbstractExtensionFinder implements ExtensionFinder, Plugin
      * @return true, if the extension finder checks extensions for its required plugins
      */
     public final boolean isCheckForExtensionDependencies() {
-        return checkForExtensionDependencies;
+        return Boolean.TRUE.equals(checkForExtensionDependencies);
     }
 
     /**
--- a/pf4j/src/main/java/org/pf4j/Extension.java
+++ b/pf4j/src/main/java/org/pf4j/Extension.java
@@ -35,6 +35,19 @@ public @interface Extension {
     int ordinal() default 0;
 
     /**
+     * An array of extension points, that are implemented by this extension.
+     * This explicit configuration overrides the automatic detection of extension points in the
+     * {@link org.pf4j.processor.ExtensionAnnotationProcessor}.
+     * <p>
+     * In case your extension is directly derived from an extension point this attribute is NOT required.
+     * But under certain <a href="https://github.com/pf4j/pf4j/issues/264">more complex scenarios</a> it
+     * might be useful to explicitly set the extension points for an extension.
+     *
+     * @return classes of extension points, that are implemented by this extension
+     */
+    Class<? extends ExtensionPoint>[] points() default {};
+
+    /**
      * An array of plugin IDs, that have to be available in order to load this extension.
      * The {@link AbstractExtensionFinder} won't load this extension, if these plugins are not
      * available / started at runtime.
--- a/pf4j/src/main/java/org/pf4j/Extension.java
+++ b/pf4j/src/main/java/org/pf4j/Extension.java
@@ -34,4 +34,16 @@ public @interface Extension {
 
     int ordinal() default 0;
 
+    /**
+     * An array of plugin IDs, that have to be available in order to load this extension.
+     * The {@link AbstractExtensionFinder} won't load this extension, if these plugins are not
+     * available / started at runtime.
+     * <p>
+     * Notice: This feature requires the optional <a href="https://asm.ow2.io/">ASM library</a>
+     * to be available on the applications classpath and has to be explicitly enabled via
+     * {@link AbstractExtensionFinder#setCheckForExtensionDependencies(boolean)}.
+     *
+     * @return plugin IDs, that have to be available in order to load this extension
+     */
+    String[] plugins() default {};
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### [Unreleased][unreleased]
 
 #### Fixed
+
+#### Changed
+
+#### Added
+
+#### Removed
+
+### [2.5.0] - 2018-12-12
+
+#### Fixed
 - [#248]: Plugin upgrade version order not guaranteed
 - [#250]: Method `DependencyResolver.resolve` removes dependents in check version block 
 - [#252]: `java.nio.file.FileSystemNotFoundException`
@@ -267,7 +277,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Removed
 
-[unreleased]: https://github.com/decebals/pf4j/compare/release-2.4.0...HEAD
+[unreleased]: https://github.com/decebals/pf4j/compare/release-2.5.0...HEAD
+[2.5.0]: https://github.com/decebals/pf4j/compare/release-2.4.0...release-2.5.0
 [2.4.0]: https://github.com/decebals/pf4j/compare/release-2.3.0...release-2.4.0
 [2.3.0]: https://github.com/decebals/pf4j/compare/release-2.2.0...release-2.3.0
 [2.2.0]: https://github.com/decebals/pf4j/compare/release-2.1.0...release-2.2.0
--- a/demo/app/pom.xml
+++ b/demo/app/pom.xml
@@ -15,7 +15,6 @@
 
     <properties>
         <main.class>org.pf4j.demo.Boot</main.class>
-        <slf4j.version>1.7.5</slf4j.version>
     </properties>
 
     <build>
@@ -25,9 +24,7 @@
                 <version>2.3</version>
                 <configuration>
                     <descriptors>
-                        <descriptor>
-                            src/main/assembly/assembly.xml
-                        </descriptor>
+                        <descriptor>src/main/assembly/assembly.xml</descriptor>
                     </descriptors>
                     <appendAssemblyId>false</appendAssemblyId>
                 </configuration>
--- a/pf4j/pom.xml
+++ b/pf4j/pom.xml
@@ -46,7 +46,7 @@
         <dependency>
             <groupId>org.ow2.asm</groupId>
             <artifactId>asm</artifactId>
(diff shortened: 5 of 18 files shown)


decebals:
I will take a look at this PR when I have a little free time. It requires a high attention.

pf4j/src/main/java/org/pf4j/Extension.java
```diff
@@ -47,4 +47,16 @@
...
+     * The {@link AbstractExtensionFinder} won't load this extension, if these plugins are not
+     * available / started at runtime.
+     * <p>
+     * Notice: This feature requires the optional <a href="https://asm.ow2.io/">ASM library</a>
+     * to be available on the applications classpath and has to be explicitly enabled via
+     * {@link AbstractExtensionFinder#setCheckForExtensionDependencies(boolean)}.
+     *
+     * @return plugin IDs, that have to be available in order to load this extension
+     */
+    String[] plugins() default {};
```
decebals: I don't like this attribute in the signature of Extension annotation. Can you are more detail about the goal of this attribute?

pinhead84:
I've [merged your last modifications](https://github.com/pf4j/pf4j/pull/270/commits/1aeb219234edbc465a062207713ed3b18e2e797f) into the pull request in order to fix a merge conflict. Besides this the new `ExtensionInfo` and `ExtensionVisitor` classes were [slightly updated](https://github.com/pf4j/pf4j/pull/270/commits/a922c53f177f42a6e3f56449cb18eb4e076c63c4) in order to support the new `points` attribute.

> I will take a look at this PR when I have a little free time. It requires a high attention.

Take your time. I'm looking forward to your feedback. ;) - In the meantime I'll test this pull request within my application. In case of problems I'm going to update this pull request accordingly or send you a message.

pf4j/src/main/java/org/pf4j/Extension.java
pinhead84: Unfortunately this attribute seems necessary to me in order to ignore extensions, that make use of an **unavailable plugin** at runtime. This avoids a possible `NoClassDefFoundError`'s in the `AbstractExtensionFinder`.

I've described my motivation about this attribute more precisely in my first posting below the headline "How extensions are processed for missing optional plugins".

decebals:
I want to start with my understanding of the optional plugin feature, to be sure that we talk the same language :).
When a plugin optionally depends on other plugin, the second plugin is referred to as the **optional** plugin. If the optional plugin is available then it is loaded and started normally. If it is not available, however, then no exception is thrown and no error is reported.

I don't have yet a simple practical use case for this feature so it will be nice if we found something.
The concept is very simple to understand and the API is simple too:
- mark the optional dependency in MANIFEST.MF or plugin.properties using **?** sign (the format is described below)
- add a new method `isOptional` in `PluginDependency`that returns true if a dependency is optional

Related to the new proposed `plugins` attribute from `Extension` I think that the goal is similar with the proposed `@RequireOptionalModule Annotation` from JIGSAW (see paragraph three from [1]). In the end RequireOptionalModule annotation was not included in Java 9 (I'm not 100% so sorry if my affirmation is not correct).

My impression is that this PR complicate a lot the actual PF4J code and the users that would use this feature will be very few (1%). Another aspect that I don't like is the use of ASM (a new dependency). Now, it's marked as optional in PF4J's pom.xml but the reality is that you, as application developer, don't know if a plugin developer will use another optional plugin in his plugin. So, in the end I think that ASM is not optional.

I will try to analyze this PR further, maybe we can improve the current implementation of this PR (in the end I think that they are two feature - optional dependency and extension that require optional dependency).

Any idea is very welcome.

@pinhead84 
Thanks for your detailed explanation/documentation and your hard work!

[1] http://openjdk.java.net/projects/jigsaw/doc/topics/optional.html

pinhead84:
> When a plugin optionally depends on other plugin, the second plugin is referred to as the **optional** plugin. If the optional plugin is available then it is loaded and started normally. If it is not available, however, then no exception is thrown and no error is reported.
>
> The concept is very simple to understand and the API is simple too:
> 
> * mark the optional dependency in MANIFEST.MF or plugin.properties using **?** sign (the format is described below)
> 
> * add a new method `isOptional` in `PluginDependency`that returns true if a dependency is optional

Yes, we're speaking the same language. ;) - This pull request follows your proposal (see [this commit](https://github.com/pf4j/pf4j/pull/270/commits/2f549f9164437071c60dd8a48d52257dd3e1fb85), [this commit](https://github.com/pf4j/pf4j/pull/270/commits/0ffcc350e88300359fc306e269377c607c9e3593) and [this commit](https://github.com/pf4j/pf4j/pull/270/commits/9fd2016be4cf8534ff81d13bca65bf2d0d6dafe1)).

> I don't have yet a simple practical use case for this feature so it will be nice if we found something.

The use case is documented in issue #266. Please let me know, if you have any questions about it or if I've made wrong assumptions.

To be honest, I can't really believe, that this is such a rare use case. I'm developing a quite simple modular Swing application without any exotic requirements / features. But it's necessary, that a plugin might add functionality to another plugin without having a forced dependency. 

I'm quite sure, that other developers with similar goals would come to similar solutions. Maybe those developers have decided not to use PF4J just because those features were missing. ;)

---

So far we're speaking the same language. But unfortunately the optional dependencies will lead to new problems, as soon as a developer likes to make use of this new feature.

PF4J is not able to load an extension class, that depends on an optional and unavailable plugin. Without the proposed `plugins` attribute in the `@Extension` annotation, PF4J simply doesn't know, if an extension is loadable or not. It could try a `Class.forName()`, but this would throw a `NoClassDefFoundError`. I guess you also dislike something like:

```java
try {
  Class extensionClass = Class.forName(extensionClassName);
  //...
} catch (NoClassDefFoundError e) {
  //... ignore the error ...
}
```

This would be a much simpler solution - but looks extremely ugly to me.

From my point of view the only clean solution would be, if PF4J knows about required plugins of an extension. PF4J can check the available plugins against the extension's requirements before the extension class is actually loaded. 

The best way to set those information is the `@Extension` annotation, because alternatives like XML for example would be pretty terrible. ;) - But let me know, if you have a better idea about it.

**Just to make sure:** Only in the specific case, that an extension needs access to an **optional** plugin, the proposed `plugins` attribute has to be used in the `@Extension`. Plugin **developers don't need to change their annotations** as long as they don't make use of optional plugin dependencies.

> Related to the new proposed `plugins` attribute from `Extension` I think that the goal is similar with the proposed `@RequireOptionalModule Annotation` from JIGSAW (see paragraph three from [1]). In the end RequireOptionalModule annotation was not included in Java 9 (I'm not 100% so sorry if my affirmation is not correct).

Of course the Jigsaw developers had to think about similar problems. But from my point of view, Jigsaw is not the same as PF4J. Jigsaw tries to specify a new distribution format for Java code but PF4J provides specific methods to extend an application. Therefore both projects might come to different conclusions about the necessity of certain features.

Or in other words: Just because Jigsaw developers decided against those features doesn't mean, that PF4J should also not provide them. ;)

> My impression is that this PR complicate a lot the actual PF4J code and the users that would use this feature will be very few (1%). 

I don't think, that this pull request complicates PF4J that much. If a developer doesn't like to use this feature, he can use PF4J as before. He doesn't have to change his `@Extension`'s and PF4J will behave like before.

> Another aspect that I don't like is the use of ASM (a new dependency). Now, it's marked as optional in PF4J's pom.xml but the reality is that you, as application developer, don't know if a plugin developer will use another optional plugin in his plugin. So, in the end I think that ASM is not optional.

But on the other hand - if an application does not provide the ASM library, the plugin developer won't be able to use optional plugin dependencies. It basically depends on the application developer, if he likes to allow optional plugin dependencies by providing the ASM library.

If you don't like to have an external ASM dependency, we also might integrate its code into PF4J. It might reduce to the total size a little bit, if only the required ASM classes are integrated. 

But I personally prefer to have a separate library. I guess most application developers would also prefer a separate ASM library. In case of critical updates in ASM, they can update the library by themselves instead of having to wait for PF4J to provide a fixed version of their integrated ASM code.

> I will try to analyze this PR further, maybe we can improve the current implementation of this PR 

I'm open for any questions or improvements.

> (in the end I think that they are two feature - optional dependency and extension that require optional dependency).

Both topics are strongly related to each other in my opinion. 

Optional plugin dependencies **(A)** mean, that a plugin might not be available at runtime (because the `DependencyResolver` ignores them). But this also means, that an unloadable extension might be accessed. This directly leads to required plugins for an extension **(B)**.

From my point of view, it doesn't really make sense to split both topics. Because **(B)** is a logical consequence of **(A)**. **(A)** doesn't really work without **(B)** - and **(B)** wouldn't make any sense without **(A)**.

---

In case this pull requests is getting merged, I would help to document this feature on your [documentation website](https://pf4j.org/), if you like to.

decebals:
I think that I found an elegant solution to our problem about ASM as dependency.
My idea is to not use reflection or ASM to extract information about an extension at runtime and instead this I propose you to push this information, at compile time, in the extension storage (`extensions.idx` file).
Now, the format of this file is very simple (on each line we have an extension class name).
Se below the content of `extensions.idx ` of plugin1 from demo:
```
# Generated by PF4J
org.pf4j.demo.welcome.WelcomePlugin$WelcomeGreeting
```

My idea is to add more information about each extension in extensions.idx. Candidate information are:
- extension points class names
- order
- required optional plugins

With this approach we obtain:
- no reflection at runtime (gain speed)
- no ASM dependency (gain space)
- we can generate in the future, in a easy mode, a nice documentation with all extensions available for each extension point, per plugin or per application (see https://github.com/pf4j/pf4j/issues/207) 
- simplify the code for the actual `ExtensionFinder` implementation

What do you think?

pinhead84:
> My idea is to not use reflection or ASM to extract information about an extension at runtime and instead this I propose you to push this information, at compile time, in the extension storage (`extensions.idx` file).

That's a nice idea. We could avoid ASM in this case. But I also like to point out some possible disadvantages.


### Separate solution for service provider extensions required

Your suggestion about `extensions.idx` only works with `LegacyExtensionStorage`. 

The `ServiceProviderExtensionStorage` uses `META-INF/services` with specific requirements about the file format. These files only may contain a `#` symbol to mark comments. PF4J should respect this format, because otherwise problems with other [`ServiceLoader`](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html) based implementations might occur.

During compilation the `ServiceProviderExtensionStorage` would extract the annotation from

```java
@Extension(plugins = {CalendarPlugin.ID, ContactsPlugin.ID})
public class ContactsCalendarFormExtension implements CalendarFormExtensionPoint {
   // ... some implementations ...
}
```

and would create a `META-INF/services/com.example.plugins.calendar.CalendarFormExtensionPoint` file, that doesn't conflict with the format specification:

```
# Generated by PF4J
com.example.plugins.contacts.addons.ContactsCalendarFormExtension # plugin=Calendar; plugin=Contacts
```

At application runtime the `ServiceProviderExtensionStorage` might read the extension parameters  from behind the comment character.

Basically this approach looks a bit like a hack - but it should work definitely. I personally would prefer a cleaner solutions, but maybe it's still acceptable.


### More complexity in PF4J

Those modifications make PF4J more complex.

- The `ExtensionStorage` mechanism has to be extended in order to support extension parameters. This might break backwards comparability and make it a bit more difficult to implement a custom `ExtensionStorage`.

- Extension storage (via `extensions.idx` or `META-INF/services`) is currently pretty straightforward. A plugin developer is not forced to use the `ExtensionAnnotationProcessor` at compile time. He can easily create those files for himself. 

  But with a more complex `ExtensionStorage` we would either need to provide more documentation about how to manually create those files, or we would force plugin developers to always use the `ExtensionAnnotationProcessor`. Maybe it's not optimal to disallow common compile parameters like `-proc:none`.

- Basically we would have to define / implement / document two different configuration methods (for both `ExtensionStorage`'s). The annotation processor copies an annotation setting into the storage file at compile time, which later at runtime has to be parsed again. This would make the internals of PF4J quite a bit more confusing for other developers. ;)

Those points might not be that much of a problem, but at least I wanted to point them out. What's not a problem for me, might still be an issue for others. 

In general I like PF4J to be as clean and easy to understand as possible. Therefore I'm looking for the most simple solution, that solves as many problems as possible.


### My personal conclusion

I personally prefer the ASM approach. Simply because it doesn't require that many changes within PF4J. It keeps the project as simple / easy as possible.

Why do we have to define two different configuration formats for extension storage, if we can just parse the Java byte code? - Java byte code contains all relevant information, was already validated by the compiler and is a well established / documented format. This avoids copying configuration values around and is absolutely future proof, in case more `@Extension` parameters are introduced with PF4J.

From my point of view, Java byte code is the format we really should look for. ;)

I personally don't mind the ASM dependency that much. It is lightweight (130KB), unintrusive and does not create any licensing problems to an application developer. Also it's not strictly necessary to use this dependency at all.

But I can fully understand, if you see this differently and don't want to depend on ASM. Of course everybody weights the pros and cons of certain solutions differently.

Please let me know, If you want to make use of ASM or prefer the discussed modifications in `LegacyExtensionStorage` and `ServiceProviderExtensionStorage`, I would add the necessary changes to my pull request in the next few days.

pinhead84:
For the sake of completeness I'd like say a word about your projected advantages:

> With this approach we obtain:
>
> no reflection at runtime (gain speed)

Currently ASM is only called once at application startup. Its results are cached into a local variable. Therefore it doesn't have any negative impact after the plugin manager was started / loaded.

Bytecode parsing with ASM is event-based (similar to SAX compared to DOM). Therefore it only parses the `@Extension` annotation and doesn't load any more data from the class file. ASM is quite stable and pretty fast (it's even part of the internal JDK API). I'm sure my current implementation with ASM also has some room for optimization.

My application currently uses round about 120 extensions and I didn't notice a negative impact with startup times regarding ASM. Of course ASM might be a bit slower compared to your suggestion, but I don't think, that performance might be a big disadvantage with ASM.

---

> With this approach we obtain:
>
> no ASM dependency (gain space)

I guess, that 113KB are not that much of a problem for most Java applications. ;) - Especially if we keep the ASM dependency optional, it might even not require any additional disk space.

---

> With this approach we obtain:
>
> we can generate in the future, in a easy mode, a nice documentation with all extensions available for each extension point, per plugin or per application (see #207)

I don't see, why those automatic documentation shouldn't be possible with ASM too.

---

> With this approach we obtain:
>
> simplify the code for the actual `ExtensionFinder` implementation

The `ExtensionFinder` might be simplified a bit on one hand. But as stated in my previous message, `ExtensionStorage` would  become more complex on the other hand.

decebals:
In the end I am OK with the solution proposed by you (your effort was decisive), the solution based on ASM. Maybe it's not the best but we have something ready to use. If you have other modifications to add before I merged the code, please do it now.

I have only one question about the new  `plugins` attribute from `Extension`. What is the semantic of this field? Is it require "optional plugins" or just `require plugins` (all nedded plugins optional or not)? 
In your use case, you have a jar that is added as dependency in a plugin - contact(s) plugin - (please correct me if it's wrong). So the extension belongs to a plugin, and you need to sppecify that the extension is actrive/enable/functional if the (optional) plugins listed in `plugins` attribute of `Extension` are loaded in application, too. Is my statement correct ?

pinhead84:
> In the end I am OK with the solution proposed by you (your effort was decisive), the solution based on ASM. Maybe it's not the best but we have something ready to use.

Thanks for your positive feedback. I also think, that it might not be the optimal approach. Any improvements are welcome. ;)

Some general thoughts on this: If PF4J wants to avoid external XML config files (as your documentation quotes "No XML, only Java."), there must be a certain price to pay. There must be a reason, why IntelliJ / Netbeans / Eclipse are still using XML.

I very much like PF4J's approach without XML. As long as configurations are kept inside of an annotation, ASM might always be a helpful tool for certain tasks and future features.

> If you have other modifications to add before I merged the code, please do it now.

I've discovered a problem in my pull request regarding automatic detection of optional plugin dependencies. After [this fix](https://github.com/pf4j/pf4j/pull/270/commits/561a994ec57fd8d158ff0369040e68ed8ba54a15) to the `AbstractExtensionFinder` the initialization works as expected.

From my point of view, there are no more fixes required. My application already uses this feature and I didn't see any problems yet. Of course I'll let you know, if I find any problems related to this feature.

You can merge this pull request as soon as you like.

> I have only one question about the new `plugins` attribute from `Extension`. What is the semantic of this field? Is it require "optional plugins" or just `require plugins` (all nedded plugins optional or not)?

The parameter in `@Extension(plugins = {CalendarPlugin.ID, ContactsPlugin.ID})` basically says, that the defined plugins

- have to be **available** at runtime (loaded into the plugin manager) and
- have to be **started** by the plugin manager at runtime

The `AbstractExtensionFinder` checks these requirements before an extension is loaded.

> In your use case, you have a jar that is added as dependency in a plugin - contact(s) plugin - (please correct me if it's wrong). So the extension belongs to a plugin, and you need to sppecify that the extension is actrive/enable/functional if the (optional) plugins listed in `plugins` attribute of `Extension` are loaded in application, too. Is my statement correct ?

Yes, that's what I'm trying to achieve.

pinhead84:
Maybe I've found an issue in the [`stopPlugin(String pluginId, boolean stopDependents)`](https://github.com/pf4j/pf4j/blob/01d592622ec74bc6cd921cb5b2b6f7f35d6939db/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L419) and [`unloadPlugin(String pluginId, boolean unloadDependents)`](https://github.com/pf4j/pf4j/blob/01d592622ec74bc6cd921cb5b2b6f7f35d6939db/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L244) methods ot `AbstractPluginManager`.

If a plugin is stopped / unloaded, the plugin manager shouldn't stop / unload **optional** dependents automatically. 

According to my example in issue #266: If I stop the calendar plugin, the contacts plugin shouldn't stop automatically. Because the contacts plugin only has an optional dependency to the calendar plugin and also works without the calendar plugin.

On first sight it looks to me, that this currently happens. But I need to investigate a bit deeper.

What do you think about this? - Maybe you shouldn't merge this pull request until we have a solution for this?

pinhead84:
> Maybe I've found an issue...

I was wrong. After further investigation I've found a bug in my application (not within PF4J), that caused this problem.

PF4J correctly ignores optional plugin dependencies in the [`stopPlugin(String pluginId, boolean stopDependents)`](https://github.com/pf4j/pf4j/blob/01d592622ec74bc6cd921cb5b2b6f7f35d6939db/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L419) and [`unloadPlugin(String pluginId, boolean unloadDependents)`](https://github.com/pf4j/pf4j/blob/01d592622ec74bc6cd921cb5b2b6f7f35d6939db/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L244) methods ot `AbstractPluginManager`.

pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
```diff
@@ -35,6 +37,8 @@
 
     protected PluginManager pluginManager;
     protected volatile Map<String, Set<String>> entries; // cache by pluginId
+    protected volatile Map<String, ExtensionInfo> extensionInfos; // cache extension infos by class name
+    protected Boolean checkForExtensionDependencies = null;
```
decebals: Why not simple `boolean`? We need three state  (false, true, null) here?
I see that `null` check is made in https://github.com/pf4j/pf4j/pull/270/files#diff-1defd675a225b032b73ef2abcb518fd2R235 but I don't know if a simple `false` check is not enough.

decebals: I understand that this flag is used to activate explicitly the extensions that require (optional) plugins feature. This flag is activated implicitly by PF4J if the plugin that contains the extension declares some optional plugins.
**Example** 
I have plugin `P1` that declares plugin `P2` as a optional dependency (via `MANIFEST.MF` or `plugin.properties`). In plugin `P1` I have an extension `E1` that requires optional plugin `P2`. To specify this thing, I will add `P2` in the `plugins` attribute of `E1` @Extension annotation.

pf4j/src/main/java/org/pf4j/Extension.java
decebals: Is it not more clear to rename `plugins` attribute to `requireOptionalPlugins`? It's a little bit long as name but it reflects the semantic.

pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
pinhead84: Basically the `checkForExtensionDependencies` has three possible states. If `checkForExtensionDependencies` is `null`, the automatic / implicit detection of optional plugin is executed.

But maybe the developer don't like to use this automatic detection at all. Therefore he might call `setCheckForExtensionDependencies` after he has created his `AbstractExtensionFinder` instance. In this case, `checkForExtensionDependencies` is either `true` or `false` and the automatic detection is disabled.

Or in other words: If `checkForExtensionDependencies` is `null`, it hasn't been set explicitly by the developer and the automatic / implicit detection of optional plugin is executed.

pf4j/src/main/java/org/pf4j/Extension.java
pinhead84: My initial idea was, to call the `points` attribute `provides` (introduced with pull request #265) and to call the `plugins` attribute `requires`. According to my personal taste

```java
@Extension(
  provides = ContactsFormExtensionPoint.class, 
  requires = {CalendarPlugin.ID, ContactsPlugin.ID}
) 
```

looks most natural to me. It's kind of short and it's more or less self-explanatory. If a developer starts using PF4J, he would have to read the documentation anyway. I guess a longer (but more precise) attribute name wouldn't really change that.

What do you think about this naming?

> Is it not more clear to rename `plugins` attribute to `requireOptionalPlugins`? It's a little bit long as name but it reflects the semantic.

In general this attribute can also be used for non-optional plugins. Maybe in the future it might make sense to extend this feature for a particular version check. An extension might depend on another / later plugin version compared to the plugin version defined via `MANIFEST.MF` / `plugin.properties`.

In case of future improvements on this feature, a name like `requireOptionalPlugins` might be misleading or had to be changed afterwards (which would break backwards compatibility). Therefore I would prefer a more general name like `plugins` or `requires`.

---

But I don't insist on a particular name for this attribute. I'm fine, if you prefer `requireOptionalPlugins` over `plugins` or `requires`. Let me know, which naming you prefer and I'm going to update the pull request accordingly.

decebals:
Thanks! Good work.

pf4j/src/main/java/org/pf4j/Extension.java
decebals: I preserved `plugins` name because is small and maybe in the future we will inject more business around this attribute.

> In general this attribute can also be used for non-optional plugins.

Here is not clear for me. See my above [example](https://github.com/pf4j/pf4j/pull/270#discussion_r247104002), I don't think that is possible to declare in `E1` extension from plugin `P1` that it requires a plugin `P3` that is not declared as dependency (optional or not) in `P1`.

pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
decebals: In this case I think that I prefer an `enum` with three possible values, something like `NONE`, `IMPLICIT`, `EXPLICIT` (or better names). But it's my preference, nothing more, so I think that I can live with Boolean :)

pf4j/src/main/java/org/pf4j/Extension.java
pinhead84: > > In general this attribute can also be used for non-optional plugins.
> 
> Here is not clear for me. See my above [example](https://github.com/pf4j/pf4j/pull/270#discussion_r247104002), I don't think that is possible to declare in `E1` extension from plugin `P1` that it requires a plugin `P3` that is not declared as dependency (optional or not) in `P1`.

That's true. I meant something different. `P1` might have a non-optional dependency on `P2` with version 1.0. But an extension provided by `P1` might depend on `P2` with version 1.1, because the corresponding extension point was added later. If `P2` is only available in version 1.0, the extension is not loaded but `P1` is still usable.

I haven't fully thought through this idea. I also don't need this at the moment. It was only meant as an example for a possible future improvement on this feature.

PR #271, opened 2019-01-03
Update slf4j

pinhead84:
PF4J currently uses a pretty old version of SLF4J. Therefore I'm suggesting to update from version **1.7.5** to **1.7.25**. Take a look at [their news page](https://www.slf4j.org/news.html) to find out more about modifications.

My application already uses PF4J together with SLF4J and Log4J without any issues. I wouldn't expect any problem for other users. But of course feel free to test it for yourself. ;)


decebals:
Thanks!

PR #277, opened 2019-01-28
configures plugin manager to load "HowdyGreeting" using the services …

robfrank:
In the demo app, the HowdyGreeting isn't loaded because the plugin manager is misconfigured.


decebals:
Good catch. Thanks!

PR #278, opened 2019-02-13
Make the project build on Java 11

dmitry-timofeev:
Update dependencies to enable build on Java 11:
- Update Mockito and plugins for 11 support.
- Disable doclint as there are numerous Javadoc issues reported.
- Remove jdk8-build profile, as 'additionalparam' property isn't
  used in POM.
- Disable Javadoc generation on 11+, as it currently fails.

Add travis build on OpenJDK11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ jdk:
 # JDK7 is not supported anymore; https://github.com/travis-ci/travis-ci/issues/7884#issuecomment-308451879
 #    - oraclejdk7
     - oraclejdk8
+    - openjdk11
 after_success:
   - mvn clean cobertura:cobertura coveralls:report
--- a/demo/app/pom.xml
+++ b/demo/app/pom.xml
@@ -42,7 +42,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
-                <version>2.3.1</version>
                 <configuration>
                     <archive>
                         <manifest>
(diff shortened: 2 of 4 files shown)


decebals:
When you think that this PR is ready for merge, please notify. Thanks!

dmitry-timofeev:
Hi @decebals, 

The Travis build is good, but it seems that it utilizes cache for compiled classes. I think it would be good to restart it with clean cache just to be sure. Locally 11 is fine except the Javadoc problems documented in the title.

I've also added the `<pluginManagement>` section so that all child modules use the same maven-compiler-plugin.

decebals:
> I think it would be good to restart it with clean cache just to be sure.

I can do it but I don't know how.

decebals:
Thanks!

dmitry-timofeev:
> I can do it but I don't know how.

In case it's needed in the future, there is an option to clear caches for a particular PR/branch:
default

decebals:
image

Thanks for information!

decebals:
> Disable Javadoc generation on 11+, as it currently fails.

We have a problem here. I cannot release a new version on Sonatype without javadoc (sonatype verify that javadoc is present). In the end, I think that javadoc is a must.

dmitry-timofeev:
I don't remember exactly, but there definitely was some kind of a problem with javadoc + maven-javadoc-plugin + Java 11. We too build Javadocs on both JDK 8 and 11, and 3.1.0 works fine for us (with a small 11-specific configuration https://github.com/exonum/exonum-java-binding/blob/af84c17eec946acda51d79e37ffbaf6a0e0e9133/exonum-java-binding/pom.xml#L685-L693). I see PF4J uses 3.0.1 — it might be worth updating.

You mentioned you build them for Sonatype — but isn't PF4J 3 Java 8-compatible (and hence, must be built and published with JDK 8)?

PR #279, opened 2019-02-13
Document where the pluginId comes from

dmitry-timofeev:
Document that `pluginId` comes from the plugin metadata. The metadata
currently can be specified in either JAR manifest as "Plugin-Id" or in a property file as "plugin.id".


decebals:
Thanks!

PR #287, opened 2019-03-01
Check no plugin with same pluginId is loaded

### LINKED ISSUE #286: PF4J overrides already loaded plugin
If there is a loaded plugin with pluginId "foo", an attempt to load another plugin with the same id "foo" (but not necessarily same provider or version) succeeds, overriding an already loaded plugin.

As PF4J 2.x APIs use pluginId as the unique identifier of the plugin in all operations (i.e., incompatible with several plugins "foo" of different versions), such attempts must be forbidden.

dmitry-timofeev:
Check no plugin with same pluginId is loaded before loading a new plugin.

Also, update Hamcrest to [2.1](https://github.com/hamcrest/JavaHamcrest/releases/tag/v2.1).

Fixes #286

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -836,12 +836,12 @@ public abstract class AbstractPluginManager implements PluginManager {
         pluginId = pluginDescriptor.getPluginId();
         if (plugins.containsKey(pluginId)) {
             PluginWrapper loadedPlugin = getPlugin(pluginId);
-            throw new PluginException(String.format("There is an already loaded plugin (%s) "
-                    + "with the same id (%s) as the plugin at path '%s'. Simultaneous loading "
+            throw new PluginException("There is an already loaded plugin ({}) "
+                    + "with the same id ({}) as the plugin at path '{}'. Simultaneous loading "
                     + "of plugins with the same PluginId is not currently supported.\n"
                     + "As a workaround you may include PluginVersion and PluginProvider "
                     + "in PluginId.",
-                loadedPlugin, pluginId, pluginPath));
+                loadedPlugin, pluginId, pluginPath);
         }
 
         log.debug("Found descriptor {}", pluginDescriptor);
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -49,17 +49,17 @@ public abstract class AbstractPluginManager implements PluginManager {
 
     private PluginDescriptorFinder pluginDescriptorFinder;
 
-    /*
+    /**
      * A map of plugins this manager is responsible for (the key is the 'pluginId').
      */
     protected Map<String, PluginWrapper> plugins;
 
-    /*
+    /**
      * A map of plugin class loaders (the key is the 'pluginId').
      */
     private Map<String, ClassLoader> pluginClassLoaders;
 
-    /*
+    /**
      * A list with unresolved plugins (unresolved dependency).
      */
     private List<PluginWrapper> unresolvedPlugins;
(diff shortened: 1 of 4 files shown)


pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -49,17 +48,17 @@
...
     private PluginDescriptorFinder pluginDescriptorFinder;
 
-    /*
+    /**
      * A map of plugins this manager is responsible for (the key is the 'pluginId').
      */
     protected Map<String, PluginWrapper> plugins;
 
-    /*
+    /**
```
decebals: My intention was to use regular block comment for private methods and fields. Private fields won’t have Javadoc generated for them unless we explicitly pass the `-private` option to the Javadoc command.
I am not against to use javadoc comments for private fields and methods but it's good to have a unitary approach throughout the project.

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -820,18 +819,31 @@ protected synchronized void firePluginStateEvent(PluginStateEvent event) {
     }
 
     protected PluginWrapper loadPluginFromPath(Path pluginPath) throws PluginException {
-        // test for plugin duplication
+        // Test for plugin path duplication
```
decebals: Start with small letter or not is not a problem for me but again it's good to have a unitary approach throughout the project.

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -820,18 +819,31 @@ protected synchronized void firePluginStateEvent(PluginStateEvent event) {
...
         log.debug("Use '{}' to find plugins descriptors", pluginDescriptorFinder);
         log.debug("Finding plugin descriptor for plugin '{}'", pluginPath);
         PluginDescriptor pluginDescriptor = pluginDescriptorFinder.find(pluginPath);
         validatePluginDescriptor(pluginDescriptor);
+
+        // Check there are no loaded plugins with the retrieved id
+        pluginId = pluginDescriptor.getPluginId();
+        if (plugins.containsKey(pluginId)) {
+            PluginWrapper loadedPlugin = getPlugin(pluginId);
+            throw new PluginException(String.format("There is an already loaded plugin (%s) "
```
decebals: Only a hint. You can write a little bit small code using [PluginException(String message, Object... args)](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/PluginException.java#L48). The format is similar with slf4j:
```java
throw new PluginException(("There is an already loaded plugin ({}) "
                    + "with the same id ({}) as the plugin at path '{}'. Simultaneous loading "
                    + "of plugins with the same PluginId is not currently supported.\n"
                    + "As a workaround you may include PluginVersion and PluginProvider "
                    + "in PluginId.",
                loadedPlugin, pluginId, pluginPath);  
```

It's my fault that `PluginException` doesn't come with a javadoc for its methods.

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
decebals: Now, I seen your comment frrom commit: 
```
Fix Javadocs for fields so that they are shown in IDE as such
```
It makes perfect sense. It's OK.

decebals:
Thanks!

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -31,6 +27,9 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.pf4j.util.StringUtils;
```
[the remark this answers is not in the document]
dmitry-timofeev: Sorry for these, I use IntelliJ. Is there a config with code style settings/an auto-formatter available?

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
dmitry-timofeev: Yes, the small benefit of these changes is that you can use "Show documentation" feature at the _use_ site to see the documentation in a pop up (Ctrl+Q/F1 depending on the configuration). That's especially useful for `protected` members (and for `private` in a big class).

Shall I submit an issue to update all of them?

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
dmitry-timofeev: Thanks, that's definitely more convenient!

I first tried to use `PluginAlreadyLoadedException`, but it has some required fields (pluginId and pluginPath) that might be ambiguous to set in this case.

PR #298, opened 2019-03-11
update asm library to version 7.1

pinhead84:
The asm library was recently updated to version 7.1. It is backwards compatible only contains bugfixes (see [changelog](https://asm.ow2.io/versions.html#7.1)). This pull requests updates the dependency version in PF4J accordingly.


decebals:
Thanks!

Issue #299, opened 2019-03-19
Provide a module-info.java


pinhead84:
@decebals: According to your message in issue #282:

> The development is done on Java 11 (OpenJDK) but the compiled classes will be compatible with Java 8.

If Java 11 is used for development / compilation, we might also consider integrating a `module-info.java`. Even if Java 8 is used as minimal / target version, we still might include this file into the build process. I've seen this for example in the [`asm-7.1.jar` file](https://repository.ow2.org/nexus/content/repositories/releases/org/ow2/asm/asm/7.1/asm-7.1.jar). The library contains a `module-info.class` (compiled for Java 9) and the rest of the class files are targeted at Java 1.5.

I like this approach, because the library is still compatible with Java 8 but can also be used as a module in Java 9+. One possible solution for this proposal is described in the [documentation of maven-compiler-plugin](https://maven.apache.org/plugins/maven-compiler-plugin/examples/module-info.html).

What do you think about this? - If you like it, I would provide a pull request for the [pf4j_3 branch](https://github.com/pf4j/pf4j/tree/pf4j_3).

decebals:
I agree with what you say. Please submit a PR.

PR #300, opened 2019-03-19
Provide a module descriptor for Java 9+

pinhead84:
This pull request addresses issue #299.

A `module-info.java` file was created. It defines the following requirements:

- [`java.base`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/module-summary.html) is always required. We don't really need to define this, but I like to set this for the sake of completeness.
- [`java.compiler`](https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/module-summary.html) is required for annotation processing.
- `org.objectweb.asm` is provided by the ASM library.
- `slf4j.api` is automatically named, because the library does not provide its own module name. Version 1.8 defines a module called `org.slf4j`. But this version is currently in beta stage. Therefore I'm not sure, if we already like to use it.
- `java.semver` is automatically named, because the library does not provide its own module name. Maybe we should send them a pull request, that at least they provide an automatic module name in their MANIFEST file. Alternatively we might integrate their source code directly into PF4J.

Currently the module exports the following packages:

- `org.pf4j`
- `org.pf4j.processor`

I'm not sure, if other exports are also necessary. Maybe we should reconsider the package hierarchy in general, that only classes are exported, which are required by 3rd party developers.

The created `pf4j.jar` file contains a subfolder `META-INF/versions/9` with the compiled `module-info.class` for Java 9+. This differs from my initial proposal #299, where I suggested to put this file into the root folder of the JAR file. But I guess this approach is a bit more reliable, because a Java 8 environment won't read this file accidentally. SLF4J in version 1.8-beta follows the same approach.

Currently I have only one little problem. According to the `.gitignore` file I guess you're also using IntelliJ IDEA? - I had to **disable automatic import of Maven projects** in the project settings and had to set **Language Level 9** for the `pf4j` module. Otherwise the IDE shows an error, if `module-info.java` is edited. Jetbrains currently offers a quite [ugly workaround](https://blog.jetbrains.com/idea/2017/10/creating-multi-release-jar-files-in-intellij-idea/), that I didn't wanted to implement due to its unnecessary complexity.

### DIFF
--- a/pf4j/pom.xml
+++ b/pf4j/pom.xml
@@ -23,7 +23,6 @@
                     <compilerArgument>-proc:none</compilerArgument>
                 </configuration>
                 <executions>
-
                     <!-- compile everything for Java 8 except the module-info.java -->
                     <execution>
                         <id>default-compile</id>
@@ -51,7 +50,6 @@
                             </includes>
                         </configuration>
                     </execution>
-
                 </executions>
             </plugin>
         </plugins>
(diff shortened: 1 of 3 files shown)


decebals:
@pinhead84 
Always when you submit an issue or a PR, it's pleasure for me to read it. You put a lot of useful details in your comments and the problems are very well described. Well done again!
Related to this subject (module descriptor for Java 9+) I can say that I have no experience. I never created a such file (module-info.java). So, I give you full credit.

> Currently I have only one little problem. According to the .gitignore file I guess you're also using IntelliJ IDEA? 

Yes, I use IDEA IntelliJ too. 

> I had to disable automatic import of Maven projects in the project settings and had to set Language Level 9 for the pf4j module. 

I do not think that's a problem for me. Thanks that you share with us this information.

PR #301, opened 2019-03-19
set version number to 3.0.0-SNAPSHOT

pinhead84:
For a better distinction between the different development branches, I would suggest to update the version numbers in the `pf4j_3` branch to `3.0.0-SNAPSHOT`.


decebals:
Good catch! Thanks!

PR #302, opened 2019-03-19
compile demo plugins for Java 8

pinhead84:
Because Java 8 is required with PF4J 3, it doesn't really make sense to compile the demo plugins for Java 7. Therefore I've updated the `pom.xml` accordingly.

In general I'm not sure, if we really need this setting anyway. Do you have any reason for providing the target Java version here again? - If we would remove this setting, it would be automatically inherited from the the root `pom.xml`, where `<maven.compiler.release>8</maven.compiler.release>` is already configured.

### DIFF
--- a/demo/plugins/pom.xml
+++ b/demo/plugins/pom.xml
@@ -16,7 +16,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <java.version>1.7</java.version>
+        <java.version>1.8</java.version>
 
         <!-- Override below properties in each plugin's pom.xml -->
         <plugin.id />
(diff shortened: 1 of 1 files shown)


decebals:
> Do you have any reason for providing the target Java version here again? 

No.

> If we would remove this setting, it would be automatically inherited from the the root pom.xml, where <maven.compiler.release>8</maven.compiler.release> is already configured.

Can you test this approach please?

pinhead84:
After removing the `<java.version>1.7</java.version>` line and the `maven-compiler-plugin` entries the demo plugins are successfully compiled for Java 8 - as specified in the root `pom.xml`.

```
% file WelcomePlugin.class 
WelcomePlugin.class: compiled Java class data, version 52.0 (Java 1.8)
```

I've updated this pull request accordingly.

decebals:
Thanks!

Issue #314, opened 2019-05-13
Remove ExtensionAnnotationProcessor log line?


cilki:
[This line](https://github.com/pf4j/pf4j/blob/6dd7a6069f0e2fbd842c81e2c8c388918b88ea81/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java#L58) in particular is showing up in my build logs all the time. None of the other log lines in `ExtensionAnnotationProcessor` are printed.

This is a really minor annoyance, but it would be nice to have the log level reduced so it doesn't print by default or have the line removed altogether.

decebals:
Please create a PR. Thanks!

decebals:
I will merge your PR, but I am curious to see your build logs to see what bothers you.

decebals:
The `info` method is used in other places too (like `process` method),.

cilki:
Here's what I get from Gradle:
```
> Task :plugin:com.sandpolis.plugin.desktop:client:mega:compileJava
Note: class org.pf4j.processor.ExtensionAnnotationProcessor init
> Task :plugin:com.sandpolis.plugin.sysinfo:client:mega:compileJava
Note: class org.pf4j.processor.ExtensionAnnotationProcessor init
> Task :plugin:com.sandpolis.plugin.sysinfo:server:vanilla:compileJava
Note: class org.pf4j.processor.ExtensionAnnotationProcessor init
...
```
I forgot that I don't actually have any extensions in these plugins yet, so that probably explains why I'm not seeing anything from `process`.

decebals:
@cilki Can we close this issue?

PR #315, opened 2019-05-13
Remove log line

cilki:
I don't see any other way to suppress this line, so I just removed it. It wasn't very informative anyway.


cilki:
Closes #315

cilki:
Closes #314

cilki:
This won't fix the whole issue because there are other log lines that will probably be emitted in the future. This needs a better solution.

PR #316, opened 2019-05-28

hazemkmammu:
unloaded plugin.

TODO: Add Unit-Test case for org.pf4j.LegacyExtensionFinder.readPluginsStorages()

### DIFF
--- a/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
@@ -20,6 +20,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


hazemkmammu:
@decebals I have replaced java.net.URLClassLoader.getResourceAsStream(String) instead of
java.net.URLClassLoader.findResources(String) in org.pf4j.LegacyExtensionFinder.readPluginsStorages(). That means only one extensions.idx file is expected for one plugin. Can there be more?

decebals:
> That means only one extensions.idx file is expected for one plugin. Can there be more?

In theory, only one `extensions.idx` file per plugin. I will double check.

It would be very nice if we can reproduce the problem with a unit test, and after this to apply this patch. 
@hazemkmammu Can you try to create a such test? Thanks!

hazemkmammu:
@decebals Is it alright to pack a tiny plugin jar (with a dummy class that does nothing) in src/test/resources to be used in the unit test?

I tried to programmatically create a plugin jar, but the class I compiled as the extension class to be packed in the jar wasn't recognized by the class loader (“ClassFormatError: Incompatible magic value”).

decebals:
@hazemkmammu 

> Is it alright to pack a tiny plugin jar (with a dummy class that does nothing) in src/test/resources to be used in the unit test?

It's OK, only to replicate the problem :) It's a very good start for me. If it's work. I can make some modifications.

Please create a separate PR for this unit test that replicates the problem.

decebals:
Replaced.

PR #317, opened 2019-05-29

hazemkmammu:
unloaded plugin.


hazemkmammu:
The test failed in the Java 7 environment because I built the dummy plugin JAR with Java 8.

decebals:
@hazemkmammu 
Thanks for your contribution. I will analyse your code and I will come with a resolution.

pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
```diff
@@ -80,12 +81,14 @@ public LegacyExtensionFinder(PluginManager pluginManager) {
             log.debug("Reading extensions storage from plugin '{}'", pluginId);
             Set<String> bucket = new HashSet<>();
 
-            try {
-                Enumeration<URL> urls = ((PluginClassLoader) plugin.getPluginClassLoader()).findResources(getExtensionsResource());
-                if (urls.hasMoreElements()) {
-                    collectExtensions(urls, bucket);
-                } else {
+            log.debug("Read '{}'", getExtensionsResource());
+            PluginClassLoader pluginClassLoader = (PluginClassLoader) plugin.getPluginClassLoader();
```
hazemkmammu: Casting to PluginClassLoader is unnecessary here.

hazemkmammu:
> @hazemkmammu
> Thanks for your contribution. I will analyse your code and I will come with a resolution.

#318 builds a plugin Jar programmatically for the test and I think is better than packing a jar in resources.

decebals:
Replaced.

Issue #282, opened 2019-02-26
PF4J 3 roadmap


decebals:
I think that is time to talk about PF4J 3.x roadmap.
PF4J is maintained under the [Semantic Versioning](https://semver.org) guidelines, so in a new major version (3.0 in our case) we can make incompatible changes in our API.
Now is time to improve PF4J without to care about backward compatible.

I will try to add some proposes:

- [x] Migrate to Java 8 (#281)
- [x] Migrate to JUnit 5 (#283)
- [ ] Async APIs (#269) (any with a PR)

The list with proposes will grow in time.

If you think that something is bad designed or you want to change something in the current API, please add a comment.

dmitry-timofeev:
Hi @decebals , thanks for putting together and leading this project!

Some possible issues to consider from a new user perspective:

1. Support providing instances of any interface as an ExtensionPoint (#289)
2. Allow to load plugin with same artifactIds (PluginIds) (or update the docs to recommend other PluginId interpretation) (#290)
3. Communicate errors with Exceptions where appropriate (#292)

decebals:
I started to work on PF4J 3.0. The modifications are available on [pf4j_3](https://github.com/pf4j/pf4j/tree/pf4j_3) branch.
The development is done on Java 11 (OpenJDK) but the compiled classes will be compatible with Java 8.

decebals:
Migrated tests to JUnit 5 with e1fb3f7bc411cdfc3eee0cacedc8f7b1a238b7cd commit.
The next step is to solve #292 (Communicate errors with Exceptions where appropriate).

pinhead84:
Maybe you can [create a milestone](https://github.com/pf4j/pf4j/milestones) for PF4J 3 and assign all related issues to it? - Maybe this would make it more transparent, which issues need to be resolved (or are already resolved) for the next major release.

decebals:
> Maybe you can create a milestone for PF4J 3 and assign all related issues to it?

The idea is good but I don't have a fixed list with all features. From this reason I cannot supply a real evolution of this work. For example, I will include #269 only if I receive a PR for this. I don't want to add more and more features only to increase the features list.

pinhead84:
> The idea is good but I don't have a fixed list with all features. From this reason I cannot supply a real evolution of this work.

I don't expect a final / fixed list of features on the milestone. Just assign all issues to the milestone, that you think are plausible / doable for the PF4J 3 release. Currently it's not very transparent, 

- which changes were already made,
[...]

Issue #217, opened 2018-05-12
Cannot remove plugin jar file after unloading


jgomer2001:
Hi, 
I'm using pf4j 2.2.0 maven dependency in my project. My web app allows an administrator user to add and drop plugins at runtime. These are the steps to "add" a plugin:

1. A .jar file is selected in the UI and uploaded to server
1. A call to `loadPlugin` is issued by passing the path where jar was stored
1. Plugin features are shown in the UI (basically the info found in the plugin descriptor)
1. The admin determines whether to onboard this plugin or reject it.

When the plugin is accepted I simply call `start` method and all is fine. When the user rejects it, I do call `unloadPlugin`, but then I need to remove the corresponding jar file. Here's what I am getting:

```
java.nio.file.FileSystemException: ...file.jar: The process has no access to the file because it's being used by another process
	...
	at java.nio.file.Files.delete(Files.java:1126) ~[?:1.8.0_65]
```

Seems very similar to [this](https://github.com/pf4j/pf4j/pull/197) which was closed months ago. 

Any idea on how to workaround it?

Thanks.

jgomer2001:
Similarly, if I already have a plugin started and make a call to `stop` and finally to `delete` , I get the same error.

decebals:
Sorry for delay. I will investigate your problem. Can you isolate the issue in a simple (quickstart) application/project? I try to reproduce the issue.

jgomer2001:
Hi, I coded  a minimalistic example, and it's embarrassing... it works perfectly

In my app I am doing some other custom manipulation of plugin jars for a variety of reasons, however it seems I need to double check if some stream is not getting closed there.  I think you may close this issue.

Thanks for your help

decebals:
I suppose that your application runs on an Windows machine. Is it correct?

jgomer2001:
yes... any tip to workaround this weird behaviour?

I skipped in my code the part where I was reading the jar file. Now the call to `Files.delete` does not complain, but the file is not deleted actually!. Only after I stop the webapp the file disappears :O

decebals:
> Hi, I coded a minimalistic example, and it's embarrassing... it works perfectly

Is it still valid this affirmation?

Do you use somehow the new jar file in your scenario (load classes from jar)?

decebals:
What servlet container (Jetty, Servlet, ...) and web framework are you using?

jgomer2001:
jetty 9.4, ZK framework

decebals:
[...]

PR #318, opened 2019-05-30

hazemkmammu:
unloaded plugin.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
@@ -20,6 +20,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.net.URL;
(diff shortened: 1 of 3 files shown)


decebals:
@hazemkmammu 
I am busy to release a new major version of PF4J. It's about [PF4J 3](https://github.com/pf4j/pf4j/issues/282).
PF4J 3 comes with a powerful abstractization of plugin packaging very useful for tests:
- [PluginJar](https://github.com/pf4j/pf4j/blob/pf4j_3/pf4j/src/test/java/org/pf4j/plugin/PluginJar.java)
- [PluginZip](https://github.com/pf4j/pf4j/blob/pf4j_3/pf4j/src/test/java/org/pf4j/plugin/PluginZip.java)

In this context I prefer that you to rewrite the test that comes with this PR, to use `PluginJar` class.
As example how to use `PluginJar` class please see [JarPluginManagerTest](https://github.com/pf4j/pf4j/blob/pf4j_3/pf4j/src/test/java/org/pf4j/JarPluginManagerTest.java). The PR must be for PF4J 3 branch. Thanks!

decebals:
@hazemkmammu 
By the way, I prefer that you add, as first step, the unit test that replicate #217.
So, I want a PR with only a unit test that replicate #217 (or something similar).
After this, you can come with a new PR (your modifications in  `LegacyExtensionFinder` and `LegacyExtensionStorage`) that fix the test fail.

hazemkmammu:
#321 recreates #217. However the test only fails in Windows.

hazemkmammu:
Replaced by #321 and #322

PR #322, opened 2019-06-04

### LINKED ISSUE #217: Cannot remove plugin jar file after unloading

hazemkmammu:
Fixes #217.


decebals:
Very good. Thanks!

Sangeeta-Yelgod-by:
I am using pf4j-update:2.3.0 version  running on window 10 machine. Please suggest which version of pf4j-update file delete issue is resolved

decebals:
@Sangeeta-Yelgod-by 
pf4j-update 2.3.0 uses pf4j 3.2. The modifications present in this PR are available in pf4j 3.0. So, it should not appear again. Please add more info (a stacktrace with the error, what you try to achieve, ...) to start investigation.

PR #323, opened 2019-06-12
Add IDEA classpath for Development mode

hank-cp:
According to [IDEA help](https://www.jetbrains.com/help/idea/specifying-compilation-settings.html), add default IDEA classpath in case developing in IDEA.

For reference, IDEA classpath setting should be changed as [here](https://stackoverflow.com/questions/13799061)

### DIFF
--- a/pf4j/src/main/java/org/pf4j/DevelopmentPluginClasspath.java
+++ b/pf4j/src/main/java/org/pf4j/DevelopmentPluginClasspath.java
@@ -41,14 +41,22 @@ public class DevelopmentPluginClasspath extends PluginClasspath {
      */
     public static final PluginClasspath KOTLIN = new PluginClasspath().addClassesDirectories("build/classes/kotlin/main", "build/resources/main", "build/tmp/kapt3/classes/main");
 
+    /**
+     * The development plugin classpath for <a href="https://www.jetbrains.com/help/idea/specifying-compilation-settings.html">IDEA</a>.
+     * The classes directories are {@code out/production/classes", out/production/resource}.
+     */
+    public static final PluginClasspath IDEA = new PluginClasspath().addClassesDirectories("out/production/classes", "out/production/resource");
+
     public DevelopmentPluginClasspath() {
         addClassesDirectories(MAVEN.getClassesDirectories());
         addClassesDirectories(GRADLE.getClassesDirectories());
         addClassesDirectories(KOTLIN.getClassesDirectories());
+        addClassesDirectories(IDEA.getClassesDirectories());
 
         addLibDirectories(MAVEN.getLibDirectories());
         addLibDirectories(GRADLE.getLibDirectories());
         addLibDirectories(KOTLIN.getLibDirectories());
+        addLibDirectories(IDEA.getLibDirectories());
     }
 
 }


decebals:
@hank-cp 
I am so sorry for the delay I answer you.
The delay is generated by the fact that, from the beginning I was not convinced by the utility of this PR. Is this PR useful for you? Do you use this setting in your projects developed with IntelliJ?
I ask you because I use IntelliJ also in my projects and I don't see how this PR will help me.
See below a test project that is created using the quickstart Maven archetype introduced by PF4J 3:
image
The structure is very similar with `demo`  included in PF4J:
- `app` maven module (contains application and API)
- `plugins` maven module that is the parent of each plugin developed for application
- `hello` and `welcome` plugins (maven modules)

I run my quickstart application via Boot class:
image
I specify that I want to run PF4J in dev mode with `-Dpf4j.mode=dev`.
In quickstart project we use Maven so we already have the information where the class files are available.
 
My specific question for you is: in what scenario (use case) the modifications proposed by you in this PR are useful? We are covered already for Maven and Gradle (build automation systems). My feeling is that don't use one of these tools and in this case is difficult to package your jars, ... (other operations related to build process).

hank-cp:
Hi @decebals 

I am using IDEA and gradle. The problem is for the main app, it will need to locate plugins' output classes folder. For IDEA, by default, it is under `out` folder (as described [here](https://www.jetbrains.com/help/idea/specifying-compilation-settings.html)), which is not covered by DevelopmentPluginClasspath. Although I could change it in IDEA project setting, but it will recover everytime I refresh from Gradle settings.

Here is my IDEA plugin module settings:
image

And how my project structure looks like:
image
* the `out` folder is generated by IDEA.
* the `build` folder is generated by Gradle.

decebals:
@hank-cp 
Thanks for explanations. I understand the problem.
I think that maybe something is wrong in your Idea config of the project.
For example, I use Maven with Idea in all my projects.
After I tell you to Idea that my project uses Maven, all the classes are generated automatically in `target\classes`. This location is taken from Maven automatically without to touch project's settings:

image
 
image
 
Again, all is made automatically by Idea IntelliJ. I don't remember to see a directory `out` (standard compiler output) in my projects. All you have to do is to specify that your project is based on Maven when you create the project in Idea IntelliJ.
I expected the same behavior for Gradle and IntelliJ.
Maybe other people that use Gradle and IntelliJ can confirm and invalidate what I tell you in this comment.

hank-cp:
Indeed, the `out` folder is generated by IDEA, while `build`/`target` is generated by Gradle or Maven. 

If I compile and launch my app by Gradle/Maven task, `build`/`target` will be there so everything is just fine. 

But when I compile and launch my app **by IDEA**, everything goes to `out`, and pf4j cannot locate class files properly, since `out` is not in its development classpath by default.

Right now I could always add `out` to development classpath in my project setting to walkaround. But I think it's good to add it to default since IDEA is getting more and more wild use.

decebals:
> But when I compile and launch my app by IDEA, everything goes to out,

What I'm trying to say is that it doesn't happen in my projects. When I specify in Idea that I use Maven in my project (when creating the project), it's smart enough not to create and use the `out` directory, and use the `target` directory (Maven standard output directory).

hank-cp:
According to this: https://stackoverflow.com/a/46035904/482533

> IntelliJ IDEA is no longer sharing the output with Gradle.

I am using IDEA 2019 BTW :)

decebals:
> According to this: https://stackoverflow.com/a/46035904/482533

Nice information. Thanks!

In this case, if you say that this PR solves your problem, I will merge it on master.

Issue #326, opened 2019-07-02
Wondering why PluginState isn't an enum


MacTrophy:
As the title states, I wonder why org.pf4j.PluginState isn't an enum? Seems to me it would help handling all cases and new values (if it ever happens) by using switch statements.

The class could easily be converted to something similar to this (based on my own enums I use in my apps)
```java
public enum PluginState {

	CREATED("CREATED"),
	DISABLED("DISABLED"),
	RESOLVED("RESOLVED"),
	STARTED("STARTED"),
	STOPPED("STOPPED");
	
	private final String status;
	
	private PluginState(String status) {
		this.status = status;
	}

	public boolean equals(String status) {
		return (status == null ? false : this.status.equalsIgnoreCase(status));
	}

	public String toString() {
		return status;
	}

	public static PluginState parse(String string) {
		for (PluginState status : values()) {
			if (status.equals(string)) {
				return status;
			}
		}

		return null;
	}
}
```

decebals:
Please create a PR that transform constants in enum, but preserve the comments from actual class (as minimal modifications as you can). Thanks!

MacTrophy:
I have zero experiences with Git (using Subversion here), can you guide me a bit on this?

I checked out the project, tried "mvn clean install" as you state in "How to build" but it failed with `DefaultPluginRepositoryTest.testGetPluginArchives:60 expected: <3> but was: <2>`. Tried adjusting this test to remove mentions of plugin 3 but now I am at `expected <2> but was <1>`. Is this test supposed to work?

I made my modifications on my local copy (using Eclipse), so far I have no compile errors but I guess running the tests would be appropriate before creating the PR?

Once I get the test running and/or you tell me to ignore it, what are the steps to provide you the PR?

decebals:
@MacTrophy 
Sure, no problem. It's time for you to learn how to create a PR (Pull Request) in GitHub :) Satisfaction will be great, I guarantee you.
You can start with https://help.github.com/en/articles/creating-a-pull-request. 
In short: 
- you must create a fork of PF4J on Github (your profile)
- bring the fork on your machine via `git clone`
- make the modifications and push them on GitHub via `git commit; git push`
- create the PR via UI from Github). 

If you need more information, please search on Google or put a comment here.

MacTrophy:
I hope I did the right thing, you should have the PR available.

decebals:
PR merged. Thanks!

PR #328, opened 2019-07-02
Modified org.pf4j.PluginState to an enum

MacTrophy:
This will help handling all cases and new values (if it ever happens) by using switch statements at least. See issue #326


decebals:
Thanks!

Issue #20, opened 2014-07-03
Advice for injecting into Plugins


gitblit:
I'm looking for advice for integrating Guice injection for Plugins and Extensions.

Here is my approach for `Extension` injection, which is working fine but you may have a better idea.

```
pf4j = new DefaultPluginManager(dir) {
  @Override
  protected ExtensionFinder createExtensionFinder() {
    DefaultExtensionFinder extensionFinder = new DefaultExtensionFinder(this) {
      @Override
      protected ExtensionFactory createExtensionFactory() {
        return new ExtensionFactory() {
          @Override
          public Object create(Class<?> extensionType) {
            logger.debug("Create instance for extension '{}'", extensionType.getName());
            try {
              return runtimeManager.getInjector().getInstance(extensionType);
            } catch (Exception e) {
              logger.error(e.getMessage(), e);
            }
            return null;
          }
        };
      }
    };
    addPluginStateListener(extensionFinder);
    return extensionFinder;
  }
};
```

Unfortunately, there is no easy way to instantiate/inject a `Plugin` without having to duplicate/rewrite large parts of pf4j. :frowning:   Or maybe I am missing something.

```
DefaultPluginManager.loadPluginDirectory->new PluginWrapper()->createPluginInstance()
```

PR #329, opened 2019-07-10
Make PluginWrapper methods public

Jotschi:
## Motivation
I'm currently faced with the problem that my custom plugin manager is unable to set the plugin factory and state since the methods are not public. The only way how I can manage this is to move my manager to the `org.pf4j` package.

## Example Usecase

My manager provides this method to quickly deploy plugins without the need to load them from filesystem:
```java
private PluginWrapper loadPlugin(Class<?> clazz, String pluginId) {
    MeshPluginDescriptor pluginDescriptor = new MeshPluginDescriptorImpl(pluginId, clazz);

    log.debug("Found descriptor {}", pluginDescriptor);
    String pluginClassName = clazz.getName();
    log.debug("Class '{}' for plugin", pluginClassName);

    // create the plugin wrapper
    log.debug("Creating wrapper for plugin '{}'", pluginClassName);
    PluginWrapper pluginWrapper = new PluginWrapper(this, pluginDescriptor, null, clazz.getClassLoader());
    pluginWrapper.setPluginFactory(getPluginFactory());

    // test for disabled plugin
    if (isPluginDisabled(pluginDescriptor.getPluginId())) {
	    log.info("Plugin '{}' is disabled", pluginClassName);
	    pluginWrapper.setPluginState(PluginState.DISABLED);
    }

    // validate the plugin
    if (!isPluginValid(pluginWrapper)) {
	    log.warn("Plugin '{}' is invalid and it will be disabled", pluginClassName);
	    pluginWrapper.setPluginState(PluginState.DISABLED);
    }

    log.debug("Created wrapper '{}' for plugin '{}'", pluginWrapper, pluginClassName);

    // add plugin to the list with plugins
    plugins.put(pluginId, pluginWrapper);
    getUnresolvedPlugins().add(pluginWrapper);

    // add plugin class loader to the list with class loaders
    getPluginClassLoaders().put(pluginId, clazz.getClassLoader());

    resolvePlugins();

    return pluginWrapper;
}
```

### DIFF
--- a/pf4j/src/main/java/org/pf4j/PluginWrapper.java
+++ b/pf4j/src/main/java/org/pf4j/PluginWrapper.java
@@ -131,11 +131,11 @@ public class PluginWrapper {
         return "PluginWrapper [descriptor=" + descriptor + ", pluginPath=" + pluginPath + "]";
     }
 
-    void setPluginState(PluginState pluginState) {
+    public void setPluginState(PluginState pluginState) {
         this.pluginState = pluginState;
     }
 
-    void setPluginFactory(PluginFactory pluginFactory) {
+    public void setPluginFactory(PluginFactory pluginFactory) {
         this.pluginFactory = pluginFactory;
     }
 


decebals:
Your custom `PluginManager` is created from scratch? I ask you because is for the first time when someone try to create a `PluginWraper`.
The actual implementation with no modifier on setter methods in `PluginWraper` is to prevent the modification from outside. Maybe a builder pattern is better.
For me it's not a problem to make public that methods (to merge this PR) but I want to understand what you try to achieve, maybe you do something wrong or a better approach is more indicated.

Jotschi:
My plugin manager extends the DefaultPluginManager and provides extra methods like the one I posted in the example.

The idea behind the whole change is to be able to quickly deploy plugins in my application. I found that the development mode of pf4j is not suitable for my usecase because plugins are run and tested in a single project.

I'm currently leading the development of a headless CMS system (https://getmesh.io/) which already has a plugin system. I'm now in the process of replacing the old system with pf4j. The CMS provides a server with a REST API which can be extended by plugins.

A plugin developer should only need to checkout the plugin project that is currently being developed.

A typical testcase in such a project looks like this:

```
        // The class rule starts a embedded in-memory version of the CMS
	@ClassRule
	public static final MeshLocalServer server = new MeshLocalServer()
		.withInMemoryMode()
		.waitForStartup();

	@Test
	public void testPlugin() {
                // A plugin can now be deployed via this mechanism. This will in turn call the custom Plugin Manager and create a new PluginWrapper / Plugin:
		Mesh mesh = server.getMesh();
		mesh.deployPlugin(BasicPlugin.class, "basic");

               // The developer can now invoke REST calls to the new plugin
	}
```

As you can see the plugin can be deployed via `mesh.deployPlugin(BasicPlugin.class, "basic");` This will in turn use the existing class loader.

This whole process is already working and I'm now in the process of polishing it and updating the existing tests.

Sidenote to the PluginManager:
I also had to override also the `getPluginFactory()` method since the mechanism with `createPluginFactory()` is not working for me due to the used dependency injection mechanism. The constructor of the `AbstractPluginManager` directly invokes `initialize()` - If I now set my injected plugin factory later on it will not be picked up by initialize. This is however currently not a problem since I'm overriding `getPluginFactory()` .

decebals:
Thanks for explanation.
I don't understand the last part of you affirmation
> I found that the development mode of pf4j is not suitable for my usecase because plugins are run and 
tested in a single project.

The part with `because plugins are run and tested in a single project` . Your plugins are run and testes in a single project or what?

I will merge this PR. I tried to understand from the PR created by you on `mesh` but it's too much new information for me.

For example this line:
```java
mesh.deployPlugin(BasicPlugin.class, "basic")
```

what is BasicPlugin.class, is it a class that extends `Plugin`? I suppose that "basic" is the plugin id. So, no plugin descriptor, loader, and other pf4j's components.

Jotschi:
Thanks for merging the PR. This will make things a lot easier.

Regarding the note on the `createPluginFactory` method:

I use dagger to handle dependency injection. Dependencies are injected via the constructor. This small example shows the problem. The injected factory can not be used by the initialize method.

```
package com.gentics.mesh.plugin.manager;

import javax.inject.Inject;
import javax.inject.Singleton;

import org.pf4j.DefaultPluginManager;
import org.pf4j.PluginFactory;

@Singleton
public class TestPluginManager extends DefaultPluginManager {

	private PluginFactory pluginFactory;

	@Inject
	public TestPluginManager(PluginFactory pluginFactory) {
		super();
		this.pluginFactory = pluginFactory;
	}

	@Override
	protected PluginFactory createPluginFactory() {
		/**
		 * pluginFactory will still be null when super constructor calls #initalize().
		 */
		return pluginFactory;
	}

}
```

But again. That's not a big problem since I can work around it.

----

> Your plugins are run and testes in a single project or what?

Yes. A typical plugin would look like this:

https://github.com/gentics/mesh-hello-world-plugin/tree/dev-plugin-revamp

All - including the test and the dev environment can be run from that project. No need to startup or checkout an extra project.


> what is BasicPlugin.class, is it a class that extends Plugin? 

`BasicPlugin.class` is a regular pf4j plugin which extends `Plugin`. And second param as you already guessed is the pluginId.

The PR in Mesh is still a bit messy. I need to clean up the PR and iron out some issues.

decebals:
Maybe help:
- https://github.com/pf4j/pf4j/issues/20
- https://github.com/gitblit/gitblit/blob/master/src/main/java/com/gitblit/manager/PluginManager.java#L121

PR #330, opened 2019-07-12
Make AbstractPluginManager fields protected

Jotschi:
This will allow custom plugin managers to add custom setters for pluginsRoot. This way the pluginsRoot can be re-configured during runtime.


Jotschi:
Thanks!

Issue #336, opened 2019-08-13
Issues with ServiceLoader used by internal JDK classes


slovdahl:
First of all, thanks for an awesome library! We've been successfully using it in production for close to 4 years now already.

We've recently stumbled upon an issue related to `ServiceLoader` used by a plugin. A little bit of context:
- the plugin uses the JAX-WS APIs available in OpenJDK 8
- the JAX-WS implementation to use is automagically discovered using Java's built-in `ServiceLoader` mechanism (we have no control over this apart from the next point)
- the implementation to use is configurable by using a `META-INF/services/javax.xml.ws.spi.Provider` file that contains the name of the class that implements the interface
- we're currently using OpenJDK 8, and OpenJDK 8 bundles a reference implementation of `javax.xml.ws.spi.Provider` (`com.sun.xml.internal.ws.spi.ProviderImpl`)
- a critical point: [CXF](https://cxf.apache.org/) is available on the class path outside the plugin, and CXF also provides an implementation of `javax.xml.ws.spi.Provider` (`org.apache.cxf.jaxws.spi.ProviderImpl`), and has a `META-INF/services/javax.xml.ws.spi.Provider` file in its JAR that points to that implementation

Our problem is that in some cases, the plugin doesn't work correctly when the CXF implementation is used, so we would like to force the plugin to use the OpenJDK bundled one instead (or the one from the `com.sun.xml.ws:jaxws-rt` Maven package when running Java 9+). This doesn't currently seem to be possible though.

After spending some time stepping through pf4j and JDK code, I think that the problem is that the `META-INF/services/javax.xml.ws.spi.Provider` file in the plugin is never found/used by `java.util.ServiceLoader`. This is caused by `ServiceLoader` using `java.lang.ClassLoader::getResources(String)`, which is a method that isn't overriden by pf4j's `PluginClassLoader`. The `java.lang.ClassLoader::getResources(String)` method will find both the file from CXF and the one in the plugin, but the one from CXF will be earlier in the returned `Enumeration<URL>`, causing it to be used.

```java
package java.lang;

public class ClassLoader {

  public Enumeration<URL> getResources(String name) throws IOException {
        @SuppressWarnings("unchecked")
        Enumeration<URL>[] tmp = (Enumeration<URL>[]) new Enumeration<?>[2];
        if (parent != null) {
            tmp[0] = parent.getResources(name);
        } else {
            tmp[0] = getBootstrapResources(name);
        }
        tmp[1] = findResources(name);

[...]

PR #337, opened 2019-08-14
Implement PluginClassLoader.getResources (#336)

slovdahl:
The logic of the method is essentially a reversed copy & paste of the
`ClassLoader::getResources(String)` method.

See #336 for a more in-depth description of the problem this PR tries to resolve.


slovdahl:
I took the liberty to write some tests for the existing `getResource(String)` method as well, mainly to validate my assumptions about how it works.

I'd be happy to add more tests if I've missed some (edge-)cases.

decebals:
It looks good.
I prefer to use [PluginZip](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/plugin/PluginZip.java) and [PluginJar](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/plugin/PluginJar.java) in tests. These classes (PluginZip and PluginJar) is not finished yet (missing support for add files - classes, jars, resources) but they can be improved relative easy.
If you have few time, please take a look to see if you can use one of these classes.
The PR is very good and it will be merged in the end. Thanks!

slovdahl:
Thank you for the feedback, and sorry for the late response.

I'm trying to figure out how to utilize `PluginZip` in the test. I have added a `PluginZip.Builder` method for adding files to the archive, but it's not obviously clear to me how e.g. unzip the archive. I see that e.g. `LoadPluginsTest` uses `PluginManager::loadPluginFromPath` for unzipping it, but it feels like using that class in the test turns it more into an integration test than unit test. What are your thoughts about it?

slovdahl:
Thanks for the rubberduck.. :duck: :smile: I think I found a solution that feels ok. I'll let the latest changes be a separate commit until you have reviewed them.

decebals:
I missed the notification that you made some changes. Sorry.
Your PR is awesome. I like the improvements to `PluginZip` (the `addFile` methods).
In the context of your second commit, do you think that we need the files added by you in `resources` in the first commit? 
I think that we can generate these files on the fly (this was the main idea when I started to implement `PluginZip` and `PluginJar`).

slovdahl:
>  In the context of your second commit, do you think that we need the files added by you in resources in the first commit?

It could work yes, but I'm not sure if it makes sense to do it via the `PluginZip` and `PluginJar` classes? The point of the files is that they should explicitly come from the parent class loader and nothing else.

slovdahl:
I removed the static files and use a static `@BeforeAll` method to create the files, please re-review.

decebals:
Good job. Thanks!

PR #339, opened 2019-10-03
Close the FileSystem that creates the path

### LINKED ISSUE #334: File handle on the plugin jar not released after stop and unload the plugin
I have a REST API that takes file inputstream of the plugin jar, load and start the plugin. I find the file handle not released after stopping and unloading the plugin. I am using pf4j 2.6.0. The link below provides a code example to reproduce the issue.
https://gist.github.com/tdong6/3f5c35f12b7fa36c9fcf8a4663e60b10

After running the test a few times, I checked `/proc/<pid_of_java_process>/fd` and found 
```
lr-x------ 1 tdong tdong 64 Jul 24 11:58 15 -> /tmp/TEST_305179773695269349/b1686424-f688-497c-a132-a6976547f082-test.jar
lr-x------ 1 tdong tdong 64 Jul 24 11:58 17 -> /tmp/TEST_653924395239229914/ee704339-dacd-422e-a8f3-7045b1f5ef95-test.jar
lr-x------ 1 tdong tdong 64 Jul 24 11:59 19 -> /tmp/TEST_7691645899945456368/5862885a-e965-470f-b0ab-0af450dcc01c-test.jar
```

tdong6:
Fix #334 by closing the FileSystem that creates the path when the file doesn't exist.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/PropertiesPluginDescriptorFinder.java
+++ b/pf4j/src/main/java/org/pf4j/PropertiesPluginDescriptorFinder.java
@@ -68,6 +68,13 @@ public class PropertiesPluginDescriptorFinder implements PluginDescriptorFinder 
 
         log.debug("Lookup plugin descriptor in '{}'", propertiesPath);
         if (Files.notExists(propertiesPath)) {
+
+            try {
+                propertiesPath.getFileSystem().close();
+            } catch (IOException e) {
+                throw new PluginException("Failed to close FileSystem that creates '{}' path", propertiesPath);
+            }
+
             throw new PluginException("Cannot find '{}' path", propertiesPath);
         }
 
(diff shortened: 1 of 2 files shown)


pf4j/src/main/java/org/pf4j/PropertiesPluginDescriptorFinder.java
```diff
@@ -68,6 +68,13 @@ protected Properties readProperties(Path pluginPath) throws PluginException {
 
         log.debug("Lookup plugin descriptor in '{}'", propertiesPath);
         if (Files.notExists(propertiesPath)) {
+
+            try {
+                propertiesPath.getFileSystem().close();
```
decebals: I am not so confident in this line.
The idea is that you don't know the type of FileSystem and probably this line doesn't work in all scenario.
Probably your case is to find `plugin.properties` in a jar/zip. It's another usual scenario (it's used always by me in my project) to run the application together with plugins from IDEA IntelliJ. In this case the PF4J try to find `plugin.properties` of each plugin in the root of each plugin:
image
In this case probably the message of exception will be `UnsuportedOperationException`.
If we want to complicate a little bit the things, we have Windows's FileSystem and Linux/Unix's File System.
So, this modification requires more tests.

tdong6: Agree. I only tested the case with jar files.

decebals:
#334 is fixed

PR #349, opened 2019-11-14
fix gradle demo

claymccoy:
This should work out of the box for the next guy:
added Gradle wrapper
fixed Gradle build errors for removed attributes
also tried to fix Gradle deprecation warnings so it won't happen again for Gradle 7, but then the build stopped acting correctly (compile -> implementation stopped libs from ending up in jars and other oddities)
added pluginsDir system property override to app Gradle build rather than needing it to pass it in
added annotation processing to Gradle build files
should use app:run rather than running the jar directly or you get a NoClassDefFoundError. This also changes the root and plugin directory needs to look in the parent dir.
updated readme with the above changes
updated Kotlin version and dependencies


decebals:
@claymccoy Thanks for your contribution!
As I said in all issues related to Gradle, I didn't used Gradle in my projects. But, my new employer uses Gradle so in the future I will be able to follow discussions about this subject.
Until now, I will merge this PR

Issue #289, opened 2019-03-01
Support any interface as an ExtensionPoint


dmitry-timofeev:
Currently a plugin cannot provide an extension implementing an interface that does _not_ extend `ExtensionPoint`. That prevents the applications from using the interfaces they do not control as extension points (e.g., from the standard library or from any third-party library).

For example, if one needs to provide an  implementation of `HashFunction` interface from Guava, they could write
```java
@Extension
public final class MyHashFunction implements HashFunction {
 …
}
```

A similar functionality is allowed in OSGi declarative services with [`@Component`](https://osgi.org/javadoc/osgi.cmpn/7.0.0/org/osgi/service/component/annotations/Component.html) annotation.

Would it be possible to support any interfaces as extension points?

pinhead84:
As far as I understand, the `@Extension` annotation is only used by the [ExtensionAnnotationProcessor](https://github.com/pf4j/pf4j/blob/3d6233154f44806b95dd4bf01e8264889492a082/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java) at compile time in order to avoid writing `META-INF/extensions.idx` or `META-INF/services` files manually. It is a convenience feature, that I don't really like to miss. ;)

Have you already tested to add an extension without this annotation to `META-INF/extensions.idx` or `META-INF/services`? - I'm not 100% sure about it. But PF4J should be able to load those extensions without the annotation, if you register these extensions manually.

decebals:
The idea with the interface marker `ExtensionPoint` is that I want:
- to have a strict control in my application/plugins, were I can "inject" extensions. The list of extensions points is defined with care in the modularization process of application. I do not think that I want any interface or abstract class to be a possible extension point.
- to have the possibility to create automatically a documentation for a plugin where to specify what extensions points and extensions provides that plugin. I think that this aspect is not covered well in actual version of PF4J and may be a subject for PF4J 3.x roadmap (https://github.com/pf4j/pf4j/issues/282). For example I think that it will be nice to have an `extension-points.idx` file, similar with `extensions.idx`, where all extensions points to be listed. Also, I think that it will be nice to include the extension point (class name) for each extension listed currently in `extensions.idx`, in this mode we will avoid the reflection in the ExtensionFinder implementations supplied by PF4J.

On the other hand, if you insist to use any interface or abstract class as an extension point, without to extends the marker interface `ExtensionPoint` (the use case from issue's description), I think that we can do it very well with what we have already.
[...]

PR #350, opened 2019-11-15
Support any interface as an ExtensionPoint

decebals:
For more information about this feature, please read #289.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
+++ b/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
@@ -52,19 +52,23 @@ import java.util.TreeSet;
 public class ExtensionAnnotationProcessor extends AbstractProcessor {
 
     private static final String STORAGE_CLASS_NAME = "pf4j.storageClassName";
+    private static final String IGNORE_EXTENSION_POINT = "pf4j.ignoreExtensionPoint";
 
     private Map<String, Set<String>> extensions = new HashMap<>(); // the key is the extension point
     private Map<String, Set<String>> oldExtensions = new HashMap<>(); // the key is the extension point
 
     private ExtensionStorage storage;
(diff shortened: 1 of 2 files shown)


decebals:
1. How it works

This feature is not enabled by default. To activate this feature you must use `pf4j.ignoreExtensionPoint` flag (option in `ExtensionAnnotationProcessor`).
If you use Maven, set this flag via compiler arguments
```xml
<plugin>
    <artifactId>maven-compiler-plugin</artifactId>
    <version>3.8.1</version>
    <configuration>
        <compilerArgs>
            <arg>-Apf4j.ignoreExtensionPoint</arg>
        </compilerArgs>
    </configuration>
</plugin>
```

2. Show me the code

The `Greeting` interface from **demo** was modified
```java
public interface Greeting /*extends ExtensionPoint*/ {

    String getGreeting();

}
```

**NOTE** Don't forget to activate this feature via `pf4j.ignoreExtensionPoint` option in `ExtensionAnnotationProcessor`.

3.  Constraints

If the class annotated with `@Extension` doesn't implement any interface
```java
@Extension
public static class WelcomeGreeting /*implements Greeting*/ {

    /*@Override*/
    public String getGreeting() {
        return "Welcome";
    }

}
```
then `ExtensionAnnotationProcessor` will throw an error when the application is compiled
```
Error:(36, 19) java: com.mycompany.welcome.WelcomePlugin.WelcomeGreeting is not an extension (it doesn't implement any interface)
```

If the class annotated with `@Extension` implements multiple interface
```java
@Extension
public static class WelcomeGreeting implements Greeting, AnInterface {

    @Override
    public String getGreeting() {
        return "Welcome";
    }

}
```
then `ExtensionAnnotationProcessor` will throw an error when the application is compiled
```
Error:(36, 19) java: com.mycompany.welcome.WelcomePlugin.WelcomeGreeting is not an extension (it implements multiple interfaces)
```

PR #352, opened 2019-12-04
adds equals/hashCode to some data classes

cfieber:
We have extended DefaultPluginDescriptor to add some extra attributes
and not having equals/hashCode implemented makes it a bit tough to
write good tests around our PluginDescriptorFinder implementation.

This just adds IntelliJ generated equals and hashCode to
DefaultPluginDescriptor, PluginDependency, and PluginClasspath.


cfieber:
I had a bit of trouble getting the project in a happy state locally to run the full test suite but hopefully travis-ci has my back ;)

Issue #355, opened 2019-12-05
Cannot remove plugin jar file after unloading,same as#217


yeahCaesar:
jgomer2001 commited issue with the same problem #217(https://github.com/jgomer2001). And I want to know , if the problem is fixed.The pf4j I'm using is 3.0.1.

yeahCaesar:
I upgrade pf4j to 3.2.0. the problem is still there

decebals:
The release notes for pf4j 3.2.0 is available [here](https://github.com/pf4j/pf4j/compare/release-3.1.0...release-3.2.0). It doesn't contain something related to #217.
I think that @jgomer2001 presented in #217 an workaround. On the other side, @hazemkmammu added a PR (##322) that fixed a `FileSystemException` in Windows on plugin delete.
Can you add more information about your problem (OS, the stacktrace of error, ...)?

yeahCaesar:
of course. OS is windows with no doubt.And the web framework is spring-boot 2.1.3.RELEASE with tomcat. Sometimes, I need to upgrade the plugin (jar file). There is a scheduler task to do the upgrade.First download the latest plugin, then stop and unload the running plugin, next is deleting the jar file, and fail to delete.File.delete( ) returns 'true', but the jar file is still there. And if you try to delete the jar file again, it thows FileAccessException.When the program is shut down, the jar disappeasr immediately.
The PluginClassLoader is still in the jvm heap after calling unload( ), is this a problem?

jgomer2001:
I faced the same as described [here](https://github.com/pf4j/pf4j/issues/217#issuecomment-391326877). Fortunately it does not show up in Linux. 

This is not a pf4j issue, but a windows jvm one IMO.

decebals:
@yeahCaesar 
Can you reproduce the problem in a minimalist/[quickstart](https://pf4j.org/dev/quickstart.html) application? A very simple application, with just one plugin.

yeahCaesar:
I think the information is enough to describe the problem.If you think a demo is necessary , I'll make one after work.

yeahCaesar:
May I ask a question? In your design , after calling the method unload , which resources should be release and which resources will be still held ?

decebals:
@yeahCaesar 
[...]

PR #356, opened 2019-12-10
release file stream when read properties, the origin code cause the p…

yeahCaesar:
…roblem that we can't delete jar file after unloading plugin with defaultPluginManager

### DIFF
--- a/pf4j/src/main/java/org/pf4j/util/FileUtils.java
+++ b/pf4j/src/main/java/org/pf4j/util/FileUtils.java
@@ -230,7 +230,9 @@ public class FileUtils {
     }
 
     public static Path getPath(URI uri, String first, String... more) throws IOException {
-        return getFileSystem(uri).getPath(first, more);
+        try (FileSystem fs = getFileSystem(uri)) {
+            return fs.getPath(first, more);
+        }
     }
 
     public static Path findFile(Path directoryPath, String fileName) {


yeahCaesar:
@decebals

decebals:
Unfortunately I cannot reproduce the problem (via `JarPluginManagerTest#deletePlugin`):
image
I think that this is because the plugin jar file from the unit test doesn't contain classes (only metadata).
Did you tested on your application that this PR fixes the problem?
I will try to see if I can find a possibility to add a unit test that replicates the problem.

decebals:
Nothing until now:
image

decebals:
Tried with a new unit test that simulate exactly your unit test described in #355 but I cannot replicate the problem:
image
image
image

So, again I cannot replicate the problem.

yeahCaesar:
image
Of course I test the modification with my local environment,  the modification resolves my problem.
As a said before, JarPluginManager is Ok , the problem occured when I use defaultPluginManager to load and start the plugin(a jar file). The defaultPluginManager tries to read properties with Proproties first while the jarPluginManager only uses the Manifest. Let me make a simple unit test to replicate the problem for you.

yeahCaesar:
` @Test
    public void deleteTest(){
        PluginManager pluginManager = new DefaultPluginManager();
        Path path = Paths.get(jarFile);
        pluginManager.loadPlugin(path);
        pluginManager.startPlugin(pluginName);
        try {
            Thread.sleep(2000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
        boolean deleted = pluginManager.deletePlugin(pluginName);
        //boolean delete = FileUtil.deleteDirHarder(path.toFile());
        System.out.println("delete result::"+deleted);
        File file = new File(jarFile);
        if(!file.exists()){
            try {
                boolean canCreateNewFile = file.createNewFile();
                assert canCreateNewFile;
            } catch (IOException e) {
                System.out.println("could not create the file which should be deleted with calling pluginManager.deletePlugin");
                assert false;
                e.printStackTrace();
            }
        }
    }`
the above code output:
deleteFail
you can see, the file.exist( ) return false, but create a file with the same name is false.Replacing DefaultPluginManager with JarPluginManage,the UT will be successful.
image
@decebals

decebals:
@yeahCaesar 
Thanks for clarification and patience!
I will try to run the same tests with `DefaultPluginManger` instead of `JarPluginManager` and I will come with news asap.

decebals:
The d359ad77ac654883db56cf9376df9b2a98c18ff9 commit added two unit tests in `DefaultPluginManager`. The `deleteJarPlugin` test replicates our problem. 
It's not clear for me why `JarManagerTest#deletePlugin` doesn't replicate the problem.

decebals:
In the end I preferred to close the `FileSystem` only for jar files, it's more safe (I am not confident that the file system should be close in any circumstances).
Please verify the patch.
Thanks for help!

Issue #359, opened 2019-12-18
Calls to File.mkdirs() in unzip logic may silently fail


slovdahl:
While debugging some issues related to module loading, I noticed that the return value of calls to `File.mkdirs()` in `org.pf4j.util.Unzip` isn't validated. The `mkdirs()` method doesn't usually throw exceptions in case the directory/directories can't be created.

I guess there can be other failures later on if there's no directory to unzip files to, but I think it would at least make sense to log something about the fact that creating directories failed.

decebals:
@slovdahl 
Can you come with a PR? Thanks!

slovdahl:
> @slovdahl
> Can you come with a PR? Thanks!

Absolutely!

What do you think would be the best course of action, just logging, or an exception? I'd be inclined to say that an exception would be warranted, because it could certainly cause other types of errors down the line if it's just logged.

decebals:
I think that you talk about [Unzip#extract](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/util/Unzip.java#L86) method.
In this case, I see that the signature of the method contains already `throws IOException` and we can throw an `IOException` if something goes wrong on `file.mkdirs`. 
Also we can remove the logger completely and the the entire `catch (FileNotFoundException e)` because of course `FileNotFoundException`  is an `IOException`.
What you say?

decebals:
The idea is to pass the responsibility of making a decision to the caller.

slovdahl:
> I think that you talk about Unzip#extract method.

Exactly, thanks for making that explicit :+1: 

>  In this case, I see that the signature of the method contains already throws IOException and we can throw an IOException if something goes wrong on file.mkdirs.
Also we can remove the logger completely and the the entire catch (FileNotFoundException e) because of course FileNotFoundException is an IOException.
What you say?

> The idea with throws IOException is to pass the responsibility of making a decision to the caller.

Very reasonable thoughts IMO. I will submit a PR with the suggested changes after the christmas holidays.

decebals:
@slovdahl 
Do you have any news?

slovdahl:
Coincidentally, I had planned on getting the MR submitted today :slightly_smiling_face: Stay tuned.

PR #362, opened 2020-01-13
Throw IOException if `mkdirs()` fails while unpacking ZIP file (#359)

slovdahl:


slovdahl:
I'm not sure if it's possible to write a test that triggers the `IOException`. Do you have any ideas @decebals?

slovdahl:
Two ideas:
- Create the plugin directory that the unzip would create, and change the owner/group of it to something that isn't writeable by the current user. But that approach has at least two problems IMO:
  1. The temporary folders created by tests can't be removed.
  1. It makes the test less portable between platforms. I'm unsure if this is a problem in practice, maybe pf4j already is fairly tied to Linux-like environments?
- Create the plugin directory that the unzip would create and a subfolder from the ZIP file, and remove the writeable/executable permission of the subfolder. This didn't work though, as the `Unzip` class does this before extracting:
  ```java
  // delete destination directory if exists
  if (destination.exists() && destination.isDirectory()) {
      FileUtils.delete(destination.toPath());
  }
  ```

pf4j/src/main/java/org/pf4j/util/Unzip.java
```diff
@@ -78,29 +78,34 @@ public void extract() throws IOException {
...
                     }
-                } catch (FileNotFoundException e) {
-                    log.error("File '{}' not found", zipEntry.getName());
                 }
             }
         }
     }
 
+    private static void mkdirsOrThrow(File dir) throws IOException {
+        if (!dir.exists()) {
```
decebals: It's not an issue but I think that it's better to put the both checks on a single line (`if (!dir.exists() && !dir.mkdirs())`)

slovdahl: Thanks, I don't know what I was thinking here :smile:

Fixed.

decebals:
Thanks!

PR #364, opened 2020-02-04
Failed plugin state added. When plugin failed to start previous state was kept

radeklos:
There is no easy way how to distinguish FAILED and STOPPED/RESOLVED plugins.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -330,9 +330,11 @@ public abstract class AbstractPluginManager implements PluginManager {
                     log.info("Start plugin '{}'", getPluginLabel(pluginWrapper.getDescriptor()));
                     pluginWrapper.getPlugin().start();
                     pluginWrapper.setPluginState(PluginState.STARTED);
+                    pluginWrapper.setFailedException(null);
                     startedPlugins.add(pluginWrapper);
                 } catch (Exception e) {
                     pluginWrapper.setPluginState(PluginState.FAILED);
+                    pluginWrapper.setFailedException(e);
                     log.error(e.getMessage(), e);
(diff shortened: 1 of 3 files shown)


pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -334,6 +334,7 @@ public void startPlugins() {
 
                     firePluginStateEvent(new PluginStateEvent(this, pluginWrapper, pluginState));
                 } catch (Exception e) {
+                    pluginWrapper.setPluginState(PluginState.FAILED);
```
decebals: I think that we need also to fire the plugin state event with:
```
firePluginStateEvent(new PluginStateEvent(this, pluginWrapper, pluginState));
```
(probably to move the call of `firePluginStateEvent` outside of the try/catch block).

Also I feel that we should keep/save the problem/exception for late use, but I don't know where:
- a property `failedException` in `PluginWrapper`
- a property (Map, ...) in `AbstractPluginManager`, something similar with `resolvedPlugins`, `startedPlugins`, ... 
- other variant

radeklos: I created the property. Not sure whether I need to clean it up somewhere else.

decebals:
Thanks!

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
decebals: > I created the property. Not sure whether I need to clean it up somewhere else.

Probably the best place is in [PluginState#setPluginState](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/PluginWrapper.java#L137) if `PluginState != FAILED` but for the moment I think that it's OK.

PR #365, opened 2020-02-07
PluginClassLoader does not resolve classpath resources from plugin dependencies

ludup:
I've been working with Pf4j and Spring and have come across an issue when using hierarchical plugins and Spring. In my project, I have a plugin that provides an SSH server component to my application. I then have several other plugins that depend on this, that provide command implementations to the SSH server.

The command implementations depend on an Interface in the SSH server plugin. It is not an ExtensionPoint, it is simply an interface and the Spring components are created by the SpringPlugin's during the creation of the ApplicationContext.

When using this setup, spring initialization fails with the current 3.3.0-SNAPSHOT because Spring attempts to load the class of my interface using getResource (I assume it does this because of all the magical stuff Spring does). This only checks the plugin and the parent class loaders, it does not consult plugin dependencies, unlike loadClass which does.

I am of course able to use and declare the interface as an ExtensionPoint and therefore not rely on Spring to create the component initially and instead inject it, but I think the fix is still valid. The fix is in the pf4j core and not pf4j-spring so its a core issue of the PluginClassLoader. If you are allowing classes to be utilized from plugin dependencies then IMHO it should be possible to access resources in those dependencies too.

This pull request contains a fix for this scenario by adding support for resources from plugin dependencies. 

It also contains a fix for an NPE that I started seeing in tests after implementing this change. The loadClassFromDependencies method checks for the nullness of classLoader, but only skips if the dependency is optional. Resulting in a NullPointerException when classLoader==null and the dependency is not optional.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/PluginClassLoader.java
+++ b/pf4j/src/main/java/org/pf4j/PluginClassLoader.java
@@ -208,8 +208,7 @@ public class PluginClassLoader extends URLClassLoader {
     }
 
     @Override
-    public Enumeration<URL> getResources(String name) throws IOException {
-        
+    public Enumeration<URL> getResources(String name) throws IOException {  
     	List<URL> resources = new ArrayList<>();
 
     	if (!parentFirst) {
@@ -288,9 +287,7 @@ public class PluginClassLoader extends URLClassLoader {
             if (classLoader == null || dependency.isOptional()) {
                 continue;
             }
-
             results.addAll(Collections.list(classLoader.findResources(name)));
-            
         }
 
         return results;
--- a/pf4j/src/main/java/org/pf4j/PluginClassLoader.java
+++ b/pf4j/src/main/java/org/pf4j/PluginClassLoader.java
@@ -24,8 +24,10 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Collection;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * One instance of this class should be created by plugin manager for every available plug-in.
@@ -175,8 +177,14 @@ public class PluginClassLoader extends URLClassLoader {
                 log.trace("Found resource '{}' in plugin classpath", name);
                 return url;
             }
+            
+            url = findResourceFromDependencies(name);
+            if (url != null) {
+                log.trace("Found resource '{}' in plugin dependencies", name);
+                return url;
+            }
 
-            log.trace("Couldn't find resource '{}' in plugin classpath. Delegating to parent", name);
+            log.trace("Couldn't find resource '{}' in plugin or dependencies classpath. Delegating to parent", name);
 
             return super.getResource(name);
         } else {
@@ -199,6 +207,8 @@ public class PluginClassLoader extends URLClassLoader {
 
             resources.addAll(Collections.list(findResources(name)));
 
+            resources.addAll(findResourcesFromDependencies(name));
+            
             if (getParent() != null) {
                 resources.addAll(Collections.list(getParent().getResources(name)));
             }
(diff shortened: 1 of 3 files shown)


decebals:
> If you are allowing classes to be utilized from plugin dependencies then IMHO it should be possible to access resources in those dependencies too.

Make sense.

I hope it does not have a side effect in applications that already use PF4J. For me it's OK.

decebals:
@ludup 
Before I will merge this PR, do you think that you can add some test(s) in [PluginClassLoaderTest](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/PluginClassLoaderTest.java)? It would be very valuable.. Thanks!

ludup:
It looks like some core changes might be required to test this or at least a complete mock PluginManager. PluginClassLoader gets dependent classloaders from the PluginManager. From what I can see the tests don't currently inject/load the plugins into the manager?

decebals:
@ludup 
Maybe a good start is to take a look on [PluginClassLoaderTest.setUp](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/PluginClassLoaderTest.java#L79).
@slovdahl did a good job with #337 and I think that he can help us.
We have already a mechanism to mock a plugin at runtime (see [PluginZip](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/plugin/PluginZip.java) and [PluginJar](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/plugin/PluginJar.java)) and we can use Mockito to mock other aspects of PF4J.
Also, I am happy to improve the testability of PF4J, to make any modifications in code that which will lead to a simpler and more concise writing of the tests.

ludup:
I was able to complete the tests just through creating a TestPluginManager that allowed me to add the PluginClassLoader for each plugin/dependency. 

I have left existing tests alone where possible, added a dependency between the plugin and parent classloader, and then added various tests to ensure only the correct files are returned for a number of scenarios including dependency only file, dependency and parent files, and files in all classloaders. 

The tests did help me discover some issues in the changes and I have fixed accordingly. These were mainly duplicate resources being returned and just required the correct call to ensure at each stage when we queried a classloader that only that classloader was queried and it did not try to resolve against its own dependencies.

I hope that all makes sense. I came up against my first real reason for needing this in my own project today, so I am looking forward to getting this merged so I can build off the latest snapshot.

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -188,25 +196,44 @@ public URL getResource(String name) {
...
+            
             return findResource(name);
         }
     }
 
     @Override
     public Enumeration<URL> getResources(String name) throws IOException {
-        if (!parentFirst) {
-            List<URL> resources = new ArrayList<>();
+
```
decebals: please remove the new lines added by this PR, from this method

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -229,5 +256,44 @@ public URL getResource(String name) {
...
+        for (PluginDependency dependency : dependencies) {
+            PluginClassLoader classLoader = (PluginClassLoader) pluginManager.getPluginClassLoader(dependency.getPluginId());
+
+            // If the dependency is marked as optional, its class loader might not be available.
+            if (classLoader == null || dependency.isOptional()) {
+                continue;
+            }
+
+            results.addAll(Collections.list(classLoader.findResources(name)));
+
```
decebals: please remove this new line

pf4j/src/test/java/org/pf4j/PluginClassLoaderTest.java
```diff
@@ -132,6 +190,7 @@ void parentLastGetResourceNonExisting() {
     void parentFirstGetResourceNonExisting() {
         assertNull(parentFirstPluginClassLoader.getResource("META-INF/non-existing-file"));
     }
+
```
decebals: please remove this new line

pf4j/src/test/java/org/pf4j/PluginClassLoaderTest.java
```diff
@@ -144,7 +203,7 @@ void parentFirstGetResourceExistsInParent() throws IOException, URISyntaxExcepti
         URL resource = parentFirstPluginClassLoader.getResource("META-INF/file-only-in-parent");
         assertFirstLine("parent", resource);
     }
-
+
```
decebals: please remove the new tab added by this PR

decebals:
The code looks good. Please fix some "issues" about formatting.

decebals:
Thanks!

decebals:
A new SNAPSHOT version is [available](https://oss.sonatype.org/content/repositories/snapshots/org/pf4j/pf4j/3.3.0-SNAPSHOT/).

PR #366, opened 2020-02-10
Check proper list for debug logging

GunoH:
I think you meant to check `result` here, instead of `entries` :-)


decebals:
Good catch. Thanks!

PR #376, opened 2020-05-25
Sort files using last modified as long

ThatGamerBlue:
This prevents downcasting errors where files can have a negative time, causing TimSort to go haywire and throw an exception.


decebals:
Thanks!

PR #383, opened 2020-07-09
parent first protected by ClassCastException instead of ClassNotFoundException

asafbennatan:


pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -133,7 +133,7 @@ public void addFile(File file) {
                 // try to load from parent
                 try {
                     return super.loadClass(className);
-                } catch (ClassCastException e) {
+                } catch (ClassNotFoundException e) {
```
decebals, line 136: I prefer only this line in PR. No formatting or changes of version.
Thanks!

PR #385, opened 2020-07-10
Better Customization for PluginClassLoader

asafbennatan:
1.added loading order to PluginClassLoader to allow loading classes in any of order from the following sources: PLUGIN,PARENT,DEPENDENCIES
2.changed several PluginClassLoader methods to protected to allow for better extension of PluginClassLoader

uses a sorted set of ClassLoadingSource to determine the correct loading order


decebals:
@asafbennatan 
I like the idea and the result.
Did you run the unit tests against these changes (PluginClassLoaderTest.java), is everything OK? 
Also I see a conflict. Can you solve this conflict and submit the result?

pf4j/src/main/java/org/pf4j/ClassLoadingSource.java
```diff
@@ -0,0 +1,5 @@
+package org.pf4j;
+
+public enum ClassLoadingSource {
```
decebals, line 3: Please add the copyright header (copy/paste from other classes) and a short description (javadoc).

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -15,38 +15,35 @@
...
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Enumeration;
-import java.util.List;
-import java.util.Objects;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import java.util.*;
```
decebals, line 25: I prefer to not use wildcard in import.

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -15,38 +15,35 @@
...
  * before delegating to the parent class loader.
- * Use {@link #parentFirst} to change the loading strategy.
+ * Use {@link #classLoadingOrder} to change the loading strategy.
  *
  * @author Decebal Suiu
  */
 public class PluginClassLoader extends URLClassLoader {
 
     private static final Logger log = LoggerFactory.getLogger(PluginClassLoader.class);
+    private static final SortedSet<ClassLoadingSource> DEFAULT_ORDER = new TreeSet<>(Arrays.asList(ClassLoadingSource.PLUGIN, ClassLoadingSource.DEPENDENCIES, ClassLoadingSource.PARENT));
```
decebals, line 38: I think that it's better to create a `ClassLoadingStrategy` and to move there these constants and to move `ClassLoadingSource` in `ClassLoadingStrategy` as a inner. 
What do you think?

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -107,135 +112,95 @@ public void addFile(File file) {
...
-                    // try next step
                 }
-
-                // look in dependencies
-                loadedClass = loadClassFromDependencies(className);
-                if (loadedClass != null) {
-                    log.trace("Found class '{}' in dependencies", className);
-                    return loadedClass;
+                if (c != null) {
+                    log.trace("Found class '{}' in " + classLoadingSource + " classpath", className);
```
decebals, line 133: Please use `log.trace("Found class '{}' in {} classpath", className, classLoadingSource);`. It's short and readable.
Also see for other `log.X` methods changed by this PR.

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
decebals: I don't think that we need a `SortedSet`. A `List` implementation (`ArrayList`) is enough.  The `ClassLoadingStrategy`  will contains a method `getSources(): Iterable`.

asafbennatan: @decebals this was to prevent the case where ClassLoadingSource is being added twice by users who use:
`    public PluginClassLoader(PluginManager pluginManager, PluginDescriptor pluginDescriptor, ClassLoader parent, SortedSet<ClassLoadingSource> classLoadingOrder) {`

asafbennatan: i think the proper name for ClassLoadingStrategy , should be something like DefaultClassLoadingStrategy to express that these are the default options and are not exhaustive ( though its quite long in name) what do you say ?

asafbennatan: also , regarding making it inner class  - since logically ClassLoadingStrategy  is actually DefaultClassLoadingStrategy  it means users using the constructor above will have to use the enums as
` DefaultClassLoadingStrategy.ClassLoadingStrategy.Plugin`
which seems a bit long , so summing up what iu think is good to do here is create a class called DefaultClassLoadingStrategy   which contains the two constants , but keep ClassLoadingStrategy as a regular non inner class

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -46,7 +47,7 @@
 
     private PluginManager pluginManager;
     private PluginDescriptor pluginDescriptor;
-    private boolean parentFirst;
+    private Set<ClassLoadingSource> classLoadingOrder;
```
decebals, line 50: Here I see only a `ClassLoadingStrategy` instance that is an `Iterable` of `ClassLoadingStrategy.Source` (enum).

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -57,11 +58,19 @@ public PluginClassLoader(PluginManager pluginManager, PluginDescriptor pluginDes
...
      */
     public PluginClassLoader(PluginManager pluginManager, PluginDescriptor pluginDescriptor, ClassLoader parent, boolean parentFirst) {
+        this(pluginManager, pluginDescriptor, parent, parentFirst ? DefaultClassLoadingStrategy.PARENT_FIRST: DefaultClassLoadingStrategy.DEFAULT_ORDER);
+    }
+
+
+    /**
+     * classloading according to {@code classLoadingOrder}
+     */
+    public PluginClassLoader(PluginManager pluginManager, PluginDescriptor pluginDescriptor, ClassLoader parent, Set<ClassLoadingSource> classLoadingOrder) {
```
decebals, line 68: The same here, pass a `ClassLoadingStrategy`

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
decebals: > i think the proper name for ClassLoadingStrategy , should be something like DefaultClassLoadingStrategy to express that these are the default options and are not exhaustive ( though its quite long in name) what do you say ?

We have only 6 possible combination no? I think that we can create all these combination and make the constructor of class private.

I think something like this:
- `A` = from Application source
- `P` = from Plugin source
- `D` = from Dependency source

Possible combination:

```
ClassLoadingStrategy.APD 
ClassLoadingStrategy.ADP
ClassLoadingStrategy.PAD
ClassLoadingStrategy.DAP
ClassLoadingStrategy.DPA
ClassLoadingStrategy.PDA
```

Do you think that is more clear?

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: done

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: done

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: done

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: i have done t the way you have suggested , the only issue that might happen is if we add another loading source for some reason we will now have 2^4 options which is not maintainable , but since i dont see another loading source ever being added here i think its ok , if for some unforeseen reason there will be another loading source ClassLoadingStrategy constructor should become public

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: done

pf4j/src/main/java/org/pf4j/ClassLoadingSource.java
asafbennatan: done

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -107,137 +115,102 @@ public void addFile(File file) {
...
-    private Class<?> loadClassFromDependencies(String className) {
+    protected Class<?> loadClassFromDependencies(String className) {
         log.trace("Search in dependencies for class '{}'", className);
         List<PluginDependency> dependencies = pluginDescriptor.getDependencies();
         for (PluginDependency dependency : dependencies) {
             ClassLoader classLoader = pluginManager.getPluginClassLoader(dependency.getPluginId());
 
             // If the dependency is marked as optional, its class loader might not be available.
-            if (classLoader == null && dependency.isOptional()) {
+            if (classLoader == null || dependency.isOptional()) {
```
decebals, line 213: I think that is not good. You revert the bug solved by #386.

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -251,38 +224,37 @@ public URL getResource(String name) {
...
-    private URL findResourceFromDependencies(String name) {
+    protected URL findResourceFromDependencies(String name) {
         log.trace("Search in dependencies for resource '{}'", name);
         List<PluginDependency> dependencies = pluginDescriptor.getDependencies();
         for (PluginDependency dependency : dependencies) {
             PluginClassLoader classLoader = (PluginClassLoader) pluginManager.getPluginClassLoader(dependency.getPluginId());
 
             // If the dependency is marked as optional, its class loader might not be available.
-            if (classLoader == null && dependency.isOptional()) {
+            if (classLoader == null || dependency.isOptional()) {
```
decebals, line 234: I think that is not good. You revert the bug solved by #386.

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -251,38 +224,37 @@ public URL getResource(String name) {
...
+    protected Collection<URL> findResourcesFromDependencies(String name) throws IOException {
         log.trace("Search in dependencies for resources '{}'", name);
         List<URL> results = new ArrayList<>();
         List<PluginDependency> dependencies = pluginDescriptor.getDependencies();
         for (PluginDependency dependency : dependencies) {
             PluginClassLoader classLoader = (PluginClassLoader) pluginManager.getPluginClassLoader(dependency.getPluginId());
 
             // If the dependency is marked as optional, its class loader might not be available.
-            if (classLoader == null && dependency.isOptional()) {
+            if (classLoader == null || dependency.isOptional()) {
```
decebals, line 255: I think that is not good. You revert the bug solved by #386.

pf4j/src/main/java/org/pf4j/ClassLoadingStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+    private ClassLoadingStrategy(List<Source> sources) {
+        this.sources = sources;
+    }
+
+    public List<Source> getSources() {
+        return sources;
+    }
+
+    public enum Source {
+        PLUGIN,PARENT,DEPENDENCIES;
```
decebals, line 67: What do you say if we rename `PLUGIN` with `APPLICATION` to be in sync with the `ClassLoadingStrategy`'s constants?

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: oops , solved

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: oops , solved

pf4j/src/main/java/org/pf4j/ClassLoadingStrategy.java
asafbennatan: done

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
asafbennatan: oops solved

decebals:
Looks good. Thanks for the contribution!

Issue #377, opened 2020-05-26
plugin.properties in 3.3.x


MacTrophy:
Has something changed between 3.2 and 3.3.1 that could have impacted the access to the `plugin.properties` at the root of JARs for the plugins? With the exact same JAR for my plugins using PF4J 3.2.0 everything works but for 3.3.0 and 3.3.1 I get lines such as
```
Lookup plugin descriptor in 'plugin.properties'
null
```
in debug verbosity level for all my plugins. This in turn makes the dependencies loading fail therefore only plugins with no dependencies will work.

decebals:
We can take a look on [Changelog](https://github.com/pf4j/pf4j/blob/master/CHANGELOG.md) document to see if we can identify a change (commit) which can have an impact.
The perfect solution is to write a unit test (JUnit) that can reproduce your problem.

MacTrophy:
The changelog doesn't contain 3.3.1 ?

decebals:
> The changelog doesn't contain 3.3.1 ?

I forgot to create a release entry for 3.3.1. It contains only a fix for a critical bug (https://github.com/pf4j/pf4j/issues/371).

decebals:
Can you create a test unit? I think that the correct place for this test is [PropertiesPluginDescriptorFinderTest](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/PropertiesPluginDescriptorFinderTest.java) and you can use [PluginJar](https://github.com/pf4j/pf4j/blob/master/pf4j/src/test/java/org/pf4j/plugin/PluginJar.java) to create the plugin jar in a dynamically mode.
This test will be excellent in the future as regression test.
In the mean time we can run the [demo](https://pf4j.org/doc/demo.html) to see if we can reproduce the problem.

decebals:
> In the mean time we can run the demo to see if we can reproduce the problem.

The demo application uses `MANIFEST.MF` as plugin descriptor container, so we need to extract the meta data about plugin from `MANIFEST.MF` into a new `plugin.properties` file.

MacTrophy:
I managed to clone the repo but the run-demo.sh fails? I have no clue on how to work with Maven.

```
[INFO] Reactor Summary for PF4J Parent 3.4.0-SNAPSHOT:
[INFO]
[INFO] PF4J Parent ........................................ SUCCESS [  1.256 s]
[INFO] PF4J ............................................... FAILURE [  0.743 s]
[...]

PR #386, opened 2020-07-16
Fix wrong logical condition (optional dependencies are always skipped)

valkuc:
Hi. There is an error in logical condition. Class loading from optional dependency is always skipped because of OR logical condition.
Also I have made some methods protected instead of private to make them usable in descendants, and added fix for starting optional dependencies. Those changes are based on use-case of pf4j library in my project. Thanks!

### DIFF
--- a/pf4j/src/main/java/org/pf4j/PluginClassLoader.java
+++ b/pf4j/src/main/java/org/pf4j/PluginClassLoader.java
@@ -15,19 +15,14 @@
  */
 package org.pf4j;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Enumeration;
-import java.util.List;
-import java.util.Objects;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import java.util.*;
 
 /**
  * One instance of this class should be created by plugin manager for every available plug-in.
@@ -177,7 +172,7 @@ public class PluginClassLoader extends URLClassLoader {
                 log.trace("Found resource '{}' in plugin classpath", name);
                 return url;
             }
-            
+
             url = findResourceFromDependencies(name);
             if (url != null) {
                 log.trace("Found resource '{}' in plugin dependencies", name);
@@ -197,41 +192,41 @@ public class PluginClassLoader extends URLClassLoader {
             log.trace("Couldn't find resource '{}' in parent", name);
 
             url = findResourceFromDependencies(name);
-           
+
             if (url != null) {
                log.trace("Found resource '{}' in dependencies", name);
                return url;
-            }  
-            
+            }
+
             return findResource(name);
         }
     }
 
     @Override
-    public Enumeration<URL> getResources(String name) throws IOException {  
+    public Enumeration<URL> getResources(String name) throws IOException {
     	List<URL> resources = new ArrayList<>();
 
     	if (!parentFirst) {
-            
+
             resources.addAll(Collections.list(findResources(name)));
 
             resources.addAll(findResourcesFromDependencies(name));
-            
+
             if (getParent() != null) {
                 resources.addAll(Collections.list(getParent().getResources(name)));
             }
 
         } else {
-        	
+
         	if (getParent() != null) {
                 resources.addAll(Collections.list(getParent().getResources(name)));
             }
-        	
+
         	resources.addAll(findResourcesFromDependencies(name));
-        	
+
         	resources.addAll(Collections.list(super.findResources(name)));
         }
-    	
+
     	return Collections.enumeration(resources);
     }
 
@@ -242,8 +237,14 @@ public class PluginClassLoader extends URLClassLoader {
             ClassLoader classLoader = pluginManager.getPluginClassLoader(dependency.getPluginId());
 
             // If the dependency is marked as optional, its class loader might not be available.
-            if (classLoader == null || dependency.isOptional()) {
-                continue;
+            if (classLoader == null) {
+                if (dependency.isOptional()) {
+                    continue;
+                } else {
+                    throw new PluginRuntimeException(
+                        String.format("Unable to load class '%s'. Required dependency '%s' not loaded.",
+                            className, dependency.getPluginId()));
+                }
             }
 
             try {
@@ -255,7 +256,7 @@ public class PluginClassLoader extends URLClassLoader {
 
         return null;
     }
-    
+
     private URL findResourceFromDependencies(String name) {
         log.trace("Search in dependencies for resource '{}'", name);
         List<PluginDependency> dependencies = pluginDescriptor.getDependencies();
@@ -263,8 +264,14 @@ public class PluginClassLoader extends URLClassLoader {
             PluginClassLoader classLoader = (PluginClassLoader) pluginManager.getPluginClassLoader(dependency.getPluginId());
 
             // If the dependency is marked as optional, its class loader might not be available.
-            if (classLoader == null || dependency.isOptional()) {
-                continue;
+            if (classLoader == null) {
+                if (dependency.isOptional()) {
+                    continue;
+                } else {
+                    throw new PluginRuntimeException(
+                        String.format("Unable to find resource '%s'. Required dependency '%s' not loaded.",
+                            name, dependency.getPluginId()));
+                }
             }
 
             URL url = classLoader.findResource(name);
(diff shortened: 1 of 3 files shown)


decebals:
@valkuc First of all I want to thank you for your interest in this project!
This PR is not simple to follow and to make a decision (merge or not).
The PR contains many modifications that are only related to formatting and from this reason is difficult to see where are the real modifications (for example I prefer to not use wildcard in imports).

I think a unit test that would reproduce the problem would have made it easier to make a decision.

Now I will try to add some comments, inline, on your modifications.

decebals:
After I analyzed your modification,  my impression (maybe I am wrong) is that you use a specific use case.
For example your wrote in one of your commit comment:
> Handle linkage error (if one plugin references class from another plugin but second one is not loaded) - otherwise loading such plugin cause application to crash

I think that us the goal of [AbstractPluginManager.resolvePlugins](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L738) method to resolve the plugins (via DependencyResolver) and to eliminate the plugins that not satisfied all requirements (missing dependencies for example). I think that is to late to solve this problem (plugin integrity) in `AbstractPluginManager.startPlugin` method.

> Make methods protected for ability to use/override them in descendant classes

Do you need all these methods to be protected?

pf4j/src/main/java/org/pf4j/PluginClassLoader.java
```diff
@@ -255,16 +256,22 @@ public URL getResource(String name) {
...
             PluginClassLoader classLoader = (PluginClassLoader) pluginManager.getPluginClassLoader(dependency.getPluginId());
 
             // If the dependency is marked as optional, its class loader might not be available.
-            if (classLoader == null || dependency.isOptional()) {
-                continue;
+            if (classLoader == null) {
+                if (dependency.isOptional()) {
+                    continue;
+                } else {
+                    throw new PluginRuntimeException(
```
decebals, line 271: This throw of `PluginRuntimeException` could break the hierarchical nature of  `PluginClassLoader` and the current behavior. So, it's a serious modification and we should analyse it very carefully.

Again, a unit test will bring more light.

Now, I don't understand very clear current lines:
```
            // If the dependency is marked as optional, its class loader might not be available.	            // If the dependency is marked as optional, its class loader might not be available.
            if (classLoader == null || dependency.isOptional()) {
```

Maybe @pinhead84 can help us (the lines have been introduced in #270).

pinhead84: Hey decebals.

> Now, I don't understand very clear current lines:
>
> ```java
> // If the dependency is marked as optional, its class loader might not be available.
> if (classLoader == null || dependency.isOptional()) {
> ```

If an optional plugin dependency is not present at runtime, the `classLoader` variable is `null`. The method should not throw an exception in this case and just continue its work. 

I'm a bit confused about the logical OR in this IF statement. In my [original commit](https://github.com/pf4j/pf4j/pull/270/commits/9fd2016be4cf8534ff81d13bca65bf2d0d6dafe1) I've used an AND. An OR doesn't really make sense to me - or I don't understand it, yet. Of course, if `classLoader != null && dependency.isOptional() == true`, the method should collect the resources instead of calling `continue`. Otherwise some resources might get ignored. Looks like a bug to me at first sight.

The IF statement has been changed with [this commit](https://github.com/pf4j/pf4j/commit/2f8343cfda32e282a9a75f7e83cd6b05b8aea93a#diff-1e3c97d1c95c106503725b202b9399bdL219).

> This throw of `PluginRuntimeException` could break the hierarchical nature of `PluginClassLoader` and the current behavior. So, it's a serious modification and we should analyse it very carefully.

I'm not a fan of those runtime exceptions. It also looks inconsistent to me as we don't throw any exception if `classLoader.loadClass(className);` fails with a `ClassNotFoundException`. Why not just do some logging here?

Best Regards.

pinhead84: > I'm a bit confused about the logical OR in this IF statement. In my [original commit](https://github.com/pf4j/pf4j/pull/270/commits/9fd2016be4cf8534ff81d13bca65bf2d0d6dafe1) I've used an AND. An OR doesn't really make sense to me - or I don't understand it, yet. Of course, if `classLoader != null && dependency.isOptional() == true`, the method should collect the resources instead of calling `continue`. Otherwise some resources might get ignored. Looks like a bug to me at first sight.

BTW: Now I've realized, that @valkuc found the problem introduced with [this commit](https://github.com/pf4j/pf4j/commit/2f8343cfda32e282a9a75f7e83cd6b05b8aea93a#diff-1e3c97d1c95c106503725b202b9399bdL219) and already fixed it - well done! The method now behaves again as it was originally intended (besides the `PluginRuntimeException`).

decebals: > I'm a bit confused about the logical OR in this IF statement.

The same feeling :smile:.

decebals: > BTW: Now I've realized, that @valkuc found the problem introduced with this commit and already fixed it - well done! The method now behaves again as it was originally intended (besides the PluginRuntimeException).

I think that:
```java
if (classLoader == null && dependency.isOptional()) {
    continue;
}
```

is enough.

valkuc: If you just replace OR with AND as it was originally then there can be NPE next in code when classloader is null and dependency is not optional. But if this situation simple can't occur in real world, then agree, using just AND will be enough.

valkuc:
> 
> 
> @valkuc First of all I want to thank you for your interest in this project!
> This PR is not simple to follow and to make a decision (merge or not).
> The PR contains many modifications that are only related to formatting and from this reason is difficult to see where are the real modifications (for example I prefer to not use wildcard in imports).
> 
> I think a unit test that would reproduce the problem would have made it easier to make a decision.
> 
> Now I will try to add some comments, inline, on your modifications.

I'm actually confused myself, but github automatically appended my next commits to this pull request. Originally I wanted to pull every single commit separatelly. But now we have what we have :) What about formatting, i'm using Idea IDE so all formatting optimization has been done by IDE. For example, it will change imports to wildcard import if there are more that 5 (default settings)

valkuc:
> 
> 
> After I analyzed your modification, my impression (maybe I am wrong) is that you use a specific use case.
> For example your wrote in one of your commit comment:
> 
> > Handle linkage error (if one plugin references class from another plugin but second one is not loaded) - otherwise loading such plugin cause application to crash
> 
> I think that us the goal of [AbstractPluginManager.resolvePlugins](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L738) method to resolve the plugins (via DependencyResolver) and to eliminate the plugins that not satisfied all requirements (missing dependencies for example). I think that is to late to solve this problem (plugin integrity) in `AbstractPluginManager.startPlugin` method.
> 

Use case is simple: I have two plugins: A, B. B optionally depends on A. For me this means that plugin B should be able to load and start even if plugin A is not available.

> > Make methods protected for ability to use/override them in descendant classes
> 
> Do you need all these methods to be protected?

This depends on your decision about merging my pull request. If you will merge it then I guess all those method will not be needed for me to override them in my subclass. If not - then at least please make `checkPluginId` protected.

decebals:
> What about formatting, i'm using Idea IDE so all formatting optimization has been done by IDE. For example, it will change imports to wildcard import if there are more that 5 (default settings)

I use Idea also. Please set wildcard import setting to 100 for this project (a high value to prevent wildcard import).

So, in order to be able to combine this PR, I ask you to do the following steps:
- remove wildcard imports
- add AND instead of OR and remove the `else` block with `PluginRuntimeException`
- remove the modifications from `PluginWrapper`
- run tests on the new PR branch to see if this code introduced a regression

valkuc:
I have a problems running unit tests on my Win10 environment. Test JarPluginManagerTest#getExtensions is failing because JUnit unable to delete directory on cleanup because file is locked (need to unload plugin but there is separate test for it). How do we solve that?

valkuc:
Another issue is that I was unable to compile project at all until I did next changes to parent pom.xml and pf4j pom.xml:

decebals:
> Another issue is that I was unable to compile project at all until I did next changes to maven-archetypes pom.xml and pf4j pom.xml:

I know about this problem. My quick fix is to use Java 11. I think that it's an issue already on this subject.

decebals:
> I have a problems running unit tests on my Win10 environment. Test JarPluginManagerTest#getExtensions is failing because JUnit unable to delete directory on cleanup because file is locked (need to unload plugin but there is separate test for it). How do we solve that?

They are few problems on Windows. For example #377. I use a Linux machine for my day to day work, so at some point, I need to install a Windows virtual VM to test these problems. Maybe people that use PF4J on Windows can help me and we can collaborate to fix them.

decebals:
@valkuc 
Now the PR looks good for me. Thanks for your contribution!

valkuc:
Thank you for nice library. I switched to PF4J from OSGi (Felix implementation) because of it simplicity and customizations ability.

PR #388, opened 2020-07-20
 Fix for #377 and minor fixes found by Sonar lint

valkuc:
Hi, I solved the issue #377 but I don't have time to write additional test for it so I checked that on Demo application. All works fine now.


decebals:
Thanks!

PR #402, opened 2020-10-14
Bucketed caching in SingletonExtensionFactory

ajithalbus:
### What 
Bucketed cache in `SingletonExtensionFactory` grouped by `ClassLoaders`

### Why
`SingletonExtensionFactory` caches extensions based on class names alone. When two extensions with same class name from two different plugins(essentially two different `ClassLoaders`) are loaded, only one of them gets instantiated and also gets served twice. This might functionally be incorrect since two extension from two different plugins should be treated different even when they share the same name.

### Quickstart 
[quickstart link](https://github.com/ajithalbus/pf4j/tree/singleton-issue-starter/singleton)
Here an additional plugin for `Greeting` called `hello-plugin-japanese`(which greets with "ohayō" which loosely translates to  "hello" in Japanese) is added along with existing `hello-plugin`. The `HelloPlugin` class exists in both `hello-plugin-japanese` and `hello-plugin` in the same package. When the plugins are loaded with `SingletonExtensionFactory`, both the plugins are been loaded successfully but `pluginManager.getExtensions(Greeting.class)` returns duplicate instances of `HelloGreeting` originating from one of the plugin only. This results in the application greeting in Japanese(or English) twice.

```
Found 4 extensions for extension point 'com.livewire.Greeting'
>>> Whazzup
>>> Welcome
>>> ohayō
>>> ohayō
```

### DIFF
--- a/pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
+++ b/pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
@@ -19,6 +19,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.WeakHashMap;
 
 /**
  * An {@link ExtensionFactory} that always returns a specific instance.
@@ -36,13 +37,12 @@ public class SingletonExtensionFactory extends DefaultExtensionFactory {
     public SingletonExtensionFactory(String... extensionClassNames) {
         this.extensionClassNames = Arrays.asList(extensionClassNames);
 
-        cache = new HashMap<>(); // simple cache implementation
+        cache = new WeakHashMap<>(); // simple cache implementation
     }
 
     @Override
     @SuppressWarnings("unchecked")
     public <T> T create(Class<T> extensionClass) {
-
         String extensionClassName = extensionClass.getName();
         ClassLoader extensionClassLoader = extensionClass.getClassLoader();
 
--- a/pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
+++ b/pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
@@ -25,12 +25,13 @@ import java.util.Map;
  * Optional, you can specify the extension classes for which you want singletons.
  *
  * @author Decebal Suiu
+ * @author Ajith Kumar
  */
 public class SingletonExtensionFactory extends DefaultExtensionFactory {
 
     private final List<String> extensionClassNames;
 
-    private Map<String, Object> cache;
+    private Map<ClassLoader, Map<String, Object>> cache;
 
     public SingletonExtensionFactory(String... extensionClassNames) {
         this.extensionClassNames = Arrays.asList(extensionClassNames);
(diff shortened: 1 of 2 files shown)


decebals:
@ajithalbus 
Thanks for this PR! For the moment, from your quickstart, I don't realize if the case described by you is a particular one or it's a common one. After I will investigate a little bit the quickstart I will come with a resolution.

decebals:
@ajithalbus 
In the end I think that the case presented by you in quickstart is a very particular case. I don't think that is possible in real life to have two (or more) extensions with the same name (package + className). I am afraid that this PR can comes with a hidden negative impact (performance, ...).
I will keep this PR open, to see if other users face the same problem. In the mean time you can create your custom `SingletonExtensionFactory` and you can inject it in your application(s).

ajithalbus:
@decebals Thanks for evaluating the PR. I agree that the case presented is quite specific and might not be a common use-case in real life. But in my opinion, I still think it's functionally incorrect for `SingletonExtensionFactory` to cache based on class names(FQN) alone given that the framework does not pose any restrictions for two extensions(from different plugins) to share the same class name. Classes with same name(FQN) loaded from different classloaders are technically **different** classes even when their implementation is identical.

pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
```diff
@@ -25,12 +25,13 @@
...
  *
  * @author Decebal Suiu
+ * @author Ajith Kumar
  */
 public class SingletonExtensionFactory extends DefaultExtensionFactory {
 
     private final List<String> extensionClassNames;
 
-    private Map<String, Object> cache;
+    private Map<ClassLoader, Map<String, Object>> cache;
```
amrith92, line 34: We can consider using a `WeakHashMap` as the backing implementation to prevent classloader instances from being skipped during GC cycles when no other references to it exists and thus preventing leaks.

amrith92:
Agree with the observation @ajithalbus made. Multiple versions of the same plugin may need to run concurrently for some use-cases, which is made possible by this patch.

pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
decebals: Replace `HashMap` with a `WeakHashMap`.

pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
```diff
@@ -41,14 +42,23 @@ public SingletonExtensionFactory(String... extensionClassNames) {
     @Override
     @SuppressWarnings("unchecked")
     public <T> T create(Class<T> extensionClass) {
+
```
decebals, line 45: Remove this empty line.

pf4j/src/test/java/org/pf4j/SingletonExtensionFactoryTest.java
```diff
@@ -43,4 +49,41 @@ public void createNewEachTime() {
         assertNotSame(extensionOne, extensionTwo);
     }
 
+    @Test
+    @SuppressWarnings("unchecked")
+    public void createNewEachTimeFromDifferentClassLoaders() throws Exception {
+
```
decebals, line 55: Remove this empty line.

pf4j/src/test/java/org/pf4j/SingletonExtensionFactoryTest.java
```diff
@@ -43,4 +49,41 @@ public void createNewEachTime() {
...
+        ClassLoader klassLoaderTwo = new URLClassLoader(classpathReferences, null);
+        Class klassTwo = klassLoaderTwo.loadClass(TestExtension.class.getName());
+
+        // create instances
+        Object instanceOne = extensionFactory.create(klassOne);
+        Object instanceTwo = extensionFactory.create(klassTwo);
+
+        // assert that instances not same
+        assertNotSame(instanceOne, instanceTwo);
+
```
decebals, line 73: Remove this empty line.

pf4j/src/test/java/org/pf4j/SingletonExtensionFactoryTest.java
```diff
@@ -43,4 +49,41 @@ public void createNewEachTime() {
...
+    }
+
+    private URL[] getClasspathReferences() throws MalformedURLException {
+
```
decebals, line 77: Remove this empty line.

decebals (changes requested):

pf4j/src/main/java/org/pf4j/SingletonExtensionFactory.java
ajithalbus: Changed to `WeakHashMap`.

decebals:
Thanks for your contribution!
I new SNAPSHOT version is available. Please [use it](https://pf4j.org/dev/maven.html) until a new release will be available.

ajithalbus:
Thank you @decebals !

PR #404, opened 2020-10-22
Support multiple plugin root directories

### LINKED ISSUE #403: Support multiple plugin root directories
It would be awesome if it was possible to configure multiple plugin root directories.

Our use-case is that we want to deliver a docker-image containing some bundled plugins, but keep the possibility to mount additional plugins via a docker-volume when starting the container.

rreich:
* all constructors now support a varargs argument instead of a single Path
* property 'pf4j.pluginsDir' now supports a comma-separated list
* default directory for enabled / disabled files is the first plugin root
* PluginManager#getPluginsRoot() returns the first plugin root

Fixes #403

I tried to  keep the changes API-compatible. If binary compatibility is necessary, we could re-add the old One-Argument-Constructors.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -25,12 +25,14 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * This class implements the boilerplate plugin code that any {@link PluginManager}
@@ -51,7 +53,7 @@ public abstract class AbstractPluginManager implements PluginManager {
     public static final String DEFAULT_PLUGINS_DIR = "plugins";
     public static final String DEVELOPMENT_PLUGINS_DIR = "../plugins";
 
-    protected Path pluginsRoot;
+    protected final List<Path> pluginsRoots = new ArrayList<>();
 
     protected ExtensionFinder extensionFinder;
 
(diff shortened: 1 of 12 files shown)


decebals:
It is not easy to make such a decision. I'll take a look and come up with news.

decebals:
To be honest I prefer to keep the actual approach with one plugins' root and you to find a solution for multiple rootss without any modification in PF4J (symbolic links, virtual file systems, ...).

rreich:
Is it that you don't like the implementation or do you think this is not a useful feature in general?

For our use-case it would be a far more elegant solution to just provide a property instead of fiddling with symlinks or even vfs.
I can't believe nobody else would find this helpful.

If it's the implementation, I have another idea:
Allow nested directory-structures in the pluginsRoot directory. If a directory does not contain a plugin (how would I test that?), (recursively) scan its subdirectories.
This way I could mount a subdirectory into the pluginsRoot from outside, even without setting any property.
It would probbably be much less invasive than the current approach.
What do you think?

I think it would be managable to extend pf4j with this feature in our project.
I will have a look at that.

decebals:
> Is it that you don't like the implementation or do you think this is not a useful feature in general?

In 8 year with PF4J it's for the first time someones asks about this functionality. So I think that feature is not a common one.
I have no problem if this feature is a bonus, something more or less hidden, without to break the current API.
When I say about "break the current API" I mean the concept part. Now we have the `pluginsRoot` concept in `PluginManager`, some `PluginRepository` implementations so it's not OK to switch to `pluginsRoots` and to return the first `Path` in the code that deals with `pluginRoot`.

The implementation proposed here definitely is not bad (it contains javadoc, some tests, good formatting, some code converted to streams).

I am curios what other users say about this feature and in particular about the switch from `pluginsRoot` to `pluginsRoots`.

fwienber:
The main use case we need this feature for is that we want to ship our application with bundled plugins, but also allow customers to add their own custom plugins. Since we package our application as a Docker container, it is easier for customers to put their plugins in a dedicated directory (Docker volume) than to add them to the same directory where the bundled plugins reside. Also, this keeps a clear separation between bundled and custom plugins. I've seen this kind of setup in other plugin-enabled applications like JetBrains' IDEs (IntelliJ IDEA et al).
Now, instead of configuring exactly two directories ("bundled-plugins" + "custom-plugins"), we thought it would make sense to generalize this to a collection of directories. A use case for a third plugin directory would be local plugin development.
Since there are options like symlinks / mount points, we don't care if these multiple plugin directories are anywhere (like in Rainer's PR) or all below a common root directory (like Rainer suggests in his response). It should just be possible to group them into different directories.

rreich:
I tried the other approach here: #406

decebals:
@fwienber Thanks for your detailed explanations!
@rreich Thanks for your effort and patience!

In the end I think that this PR, as approach. is more clean that #406.
I will submit it soon.
I want to leave a little time to those who do not find this feature useful or consider the implementation to break current API.
If everyone votes for, then this feature will be available in the next release (3.5.0).

decebals:
@rreich In the meantime please verify very well this PR (with your application(s)) and add more unit tests if it's possible.

rreich:
Thank you for deciding to add this feature!
I have tested this PR with our application now loading plugins from one and multiple directories and it works fine.
Regarding unit-tests I fixed a bug in my new test class and added multiple roots to the DefaultPluginRepositoryTest.

decebals:
@rreich Thanks!

rreich:
Thanks for Merging!

PR #406, opened 2020-10-27
Enable finding plugins in subdirectories of pluginsRoot

rreich:
Alternative approach addressing #403

DefaultPluginRepository scans for plugins in subfolders when a PluginDescriptorFinder is provided, so that it can decide if a subfolder is a plugin or just a directory possibly containing plugins or further subfolders.
To this end the AbstractPluginManager has been changed to initialize the finder first to be able to provide it to the PluginRepository creation.
There is some programming-by-exception when using the finder to decide if a directory is a plugin, but I wanted to do the exact same things as when creating the PluginDescriptor to make sure the check is really consistent with the creation.


rreich:
Closing in favor of #404

PR #412, opened 2020-11-23
Update pom.xml

l1046j:
org.slf4j.ext.EventData in the slf4j-ext module in QOS.CH SLF4J before 1.8.0-beta2 allows remote attackers to bypass intended access restrictions via crafted data.
update the version of slf4j

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.release>8</maven.compiler.release>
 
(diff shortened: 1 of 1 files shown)


decebals:
Thanks! Is it OK to go with a beta version? Is it situation so critical?

l1046j:
> Thanks! Is it OK to go with a beta version? Is it situation so critical?

I'm not sure :),but we need to use pf4j without known vulnerabilities.

decebals:
@l1046j In the end I think that I'm not comfortable with the idea to have a dependency in a beta version. Please notify me when a general release is available. Thank you for understanding!

Issue #414, opened 2020-12-31
adding the programatic plugin generation interface to a different dependency


asafbennatan:
Hi,
i think it would be beneficial if the classes used for pf4j's testing are published in a different package so framework extenders could also check their work in the same manner.

decebals:
Yes, you are right. I tried to create a `pf4j-test` repository and to move some classes from pf4j/test (like `PluginZip`, `PluginJar`, ...) in the new repo but I encountered some problems. I don't remember exactly but I think that was a circular dependency or something similar.

asafbennatan:
for my use case i only took:
DefaultClassDataProvider
PluginJar
ClassDataProvider

and used plugin jar class

asafbennatan:
actuall i understand why becuase testClasses->pf4j->testClasses

decebals:
> actually i understand why because pf4j-tests->pf4j->testClasses,
> i guess the correct way of solving that is to create three packages:

It's not clear for me if `three packages` means three repositories or three artifacts (Maven, Jar) or just three Java packages (group of classes).
Probably the simple solution is to move the important classes (the classes mentioned by you) from `src/test` to `src/main` (in a new `org.pf4j.test` package). The single drawback of this solution is that artifact (the jar file) will increase in size. In the end I don't know if it's a serious drawback.

asafbennatan:
i mean jars(maven) , while moving the test classes into main is a possibility IMHO the main issue here is the leakage of test related classes into the library regardless of the size , so my suggestions will be to create the following maven projects:
1.pf4j-api containing production classes required by testing as well as main library
2.pf4j-tests a library which will be used by testing only ,contains tests classes- this will depend on pf4j-api
3.pf4j (current library) uses pf4j-tests in test scope and pf4j-api in compile(default) scope - this contains all production classes which are not required by pf4j-tests

decebals:
To be honest I don't like the idea to split `pf4j` in `pf4j-api` and `pf4j`. It's very difficult to decide what to contain `pf4j-api`.
I prefer to create only a new `pf4j-test` Maven module in the current repository, and to move some classes from `src/test` (`pf4j` module) in the new `pf4j-test` module. Of course, if it's possible and enough.

asafbennatan:
Yes it should be enough

decebals:
In the end I found a simple and not intrusive (no need to create a new Module module or something similar).
[...]

PR #415, opened 2021-01-06
Externalize some useful classes from testing

decebals:
The idea of this PR is to allow other PF4J-based applications to use classes from `test/src`. 
We have some interesting concepts like `PluginJar`, `PluginZip`, `ClassDataProvider` (and more will probably be added in the future).

For more details see https://github.com/pf4j/pf4j/issues/414.

It's very simple to use these classes in your test suite.
Just add the snippet below to your POM file and that it's all:
```xml
<dependency>
    <groupId>org.pf4j</groupId>
    <artifactId>pf4j</artifactId>
    <version>${pf4j.version}</version>
    <!--            
    <classifier>tests</classifier>
    -->
    <type>test-jar</type>
    <scope>test</scope>
</dependency>
```

### DIFF
--- a/pf4j/pom.xml
+++ b/pf4j/pom.xml
@@ -73,7 +73,7 @@
                         </goals>
                         <configuration>
                             <includes>
-                                <inculde>org/pf4j/plugin/*</inculde>
+                                <inculde>org/pf4j/test/*</inculde>
                             </includes>
                         </configuration>
                     </execution>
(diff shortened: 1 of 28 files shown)


decebals:
I tested with an external application and the result is very good.
The external application contain only a test (`ManifestPluginDescriptorFinderTest`) that was extracted from test suite of PF4J:
image

Issue #250, opened 2018-11-21
DependencyResolver.resolve - removing dependents in check version block ?


surajpuvvada:
In `DependencyResolver.resolve()` - I noticed this block of code (https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/DependencyResolver.java#L93) where dependents.remove(0) is called. Doesn't this remove the dependent plugin ? Is this correct behavior ? 
For example: If I have A depend on B and I try to unload B it simply goes and unloads B without unloading A as `getDependents()` returns empty in `AbstractDefaultManager.unloadPlugin()` call. 

```java
// check dependencies versions
for (PluginDescriptor plugin : plugins) {
    String pluginId = plugin.getPluginId();
    String existingVersion = plugin.getVersion();

    List<String> dependents = getDependents(pluginId);
    while (!dependents.isEmpty()) {
        String dependentId = dependents.remove(0);
        PluginDescriptor dependent = pluginByIds.get(dependentId);
        String requiredVersion = getDependencyVersionSupport(dependent, pluginId);
        boolean ok = checkDependencyVersion(requiredVersion, existingVersion);
        if (!ok) {
            result.addWrongDependencyVersion(new WrongDependencyVersion(pluginId, dependentId, existingVersion, requiredVersion));
        }
    }
}
```

surajpuvvada:
Thanks for the reply. 
I think the issue is the *dependents*  list is a direct read through from the DAG i.e. not cloned. So *dependents.remove(0)* removes it. 

When I changed the code to use an iterator instead - it seems to work correctly.

`// check dependencies versions
        for (PluginDescriptor plugin : plugins) {
            String pluginId = plugin.getPluginId();
            String existingVersion = plugin.getVersion();

            List<String> dependents = getDependents(pluginId);
            Iterator<String> it = dependents.iterator();
            while(it.hasNext()){
                String dependentId = it.next();
                PluginDescriptor dependent = pluginByIds.get(dependentId);
                String requiredVersion = getDependencyVersionSupport(dependent, pluginId);
                boolean ok = checkDependencyVersion(requiredVersion, existingVersion);
                if (!ok) {
                    result.addWrongDependencyVersion(new WrongDependencyVersion(pluginId, dependentId, existingVersion, requiredVersion));
                    it.remove();
                }
            }
        }`

decebals:
@surajpuvvada 
[...]

PR #416, opened 2021-01-07
#394, keep dependent when unload plugin

### LINKED ISSUE #394: DependencyResolver lost dependent info after plugin stop
Have two plugins A and B, B depends on A.

### Step to reproduce:
1. Resolve and start A/B
2. Stop A, A/B will be both `STOPPED`, which is expected.
3. Start A.
4. Start B,
5. Stop A again, A is `STOPPED` but B remains `STARTED`, which expected to be both `STOPPED`.

### Possible cause
in `AbstractPluginManager.stopPlugin(plugin, stopDependents)` methods, the following line
```
String dependent = dependents.remove(0);
```
also change the dependencyResolver, the dependent info is dropped with plugin stopping, unless we resolve everything again.

*a reproducible repo should be found [here](https://github.com/j-barata/marketbasket-sbp-sample). The pf4j version it's using is v3.1.0, and same behavior under v3.4.1*

hank-cp:
fix #394. Not sure will this change break other place.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -274,8 +274,7 @@ public abstract class AbstractPluginManager implements PluginManager {
         try {
             if (unloadDependents) {
                 List<String> dependents = dependencyResolver.getDependents(pluginId);
-                while (!dependents.isEmpty()) {
-                    String dependent = dependents.get(0);
+                for (String dependent : dependents) {
                     unloadPlugin(dependent, false);
                     dependents.addAll(0, dependencyResolver.getDependents(dependent));
                 }
@@ -467,8 +466,7 @@ public abstract class AbstractPluginManager implements PluginManager {
 
         if (stopDependents) {
             List<String> dependents = dependencyResolver.getDependents(pluginId);
-            while (!dependents.isEmpty()) {
-                String dependent = dependents.remove(0);
+            for (String dependent : dependents) {
                 stopPlugin(dependent, false);
                 dependents.addAll(0, dependencyResolver.getDependents(dependent));
             }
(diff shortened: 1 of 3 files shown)


pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -275,7 +275,7 @@ protected boolean unloadPlugin(String pluginId, boolean unloadDependents) {
             if (unloadDependents) {
                 List<String> dependents = dependencyResolver.getDependents(pluginId);
                 while (!dependents.isEmpty()) {
-                    String dependent = dependents.remove(0);
+                    String dependent = dependents.get(0);
```
decebals, line 278: I like the change because is very light. 
Did you tested? I ask you because at first glance, this change produces an infinite loop (`dependents` will never be empty and the while block will be executed forever).

hank-cp:
ops...I didn't notice that and left it to unit test (all passed...)

leave it for a while, I will rewrite it and  add some unit test to cover the case.

hank-cp:
@decebals updated.

also add unit test: PluginDependencyTest.dependentStop

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -274,8 +274,7 @@ protected boolean unloadPlugin(String pluginId, boolean unloadDependents) {
         try {
             if (unloadDependents) {
                 List<String> dependents = dependencyResolver.getDependents(pluginId);
-                while (!dependents.isEmpty()) {
-                    String dependent = dependents.remove(0);
+                for (String dependent : dependents) {
                     unloadPlugin(dependent, false);
                     dependents.addAll(0, dependencyResolver.getDependents(dependent));
```
decebals, line 279: I think that this line will produce in the end a `java.util.ConcurrentModificationException`.

decebals:
I see something related to this issue. It's about https://github.com/pf4j/pf4j/issues/250.
Maybe it helps.

decebals:
Replaced by https://github.com/pf4j/pf4j/pull/416. Much of the work was done by by @hank-cp. Thanks!

PR #426, opened 2021-02-22
Update module-info.java

### LINKED ISSUE #405: Java 9 - module-info problems
Since using Java 9 with module-info, I cannot build the project anymore.

I can install the current master with maven on command-line, but I cannot build the project in Intellij Idea due to unknown module-info files.

When I change the line https://github.com/pf4j/pf4j/blob/master/pom.xml#L46 from version 8 to 9 (which would be the correct value IMHO), I can build in Idea, but not on the command-line anymore.

```
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project pf4j: Compilation failure: Compilation failure:
...module-info.java:[30,27] module not found: org.objectweb.asm
...module-info.java:[36,19] module not found: slf4j.api
...module-info.java:[41,18] module not found: java.semver
```
The unit-tests have strange NoClassDefFound errors and duplicate resource-loading problems.

I think something regarding the Java Modules is not set-up correctly.
Or maybe I am doing something wrong?

decebals:
Try to fix https://github.com/pf4j/pf4j/issues/405.

### DIFF
--- a/pf4j/src/main/java/module-info.java
+++ b/pf4j/src/main/java/module-info.java
@@ -26,8 +26,8 @@ module org.pf4j {
     // provides javax.annotation
     requires java.compiler;
 
-    // provided by the ASM library
-    requires org.objectweb.asm;
+    // provided by the ASM library, use "requires static" since it's optional
+    requires static org.objectweb.asm;
 
     // The SLF4J library currently does not provide a module.
     // Version 1.8 provides a module called "org.slf4j". But this version is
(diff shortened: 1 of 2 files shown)


decebals:
The remaining problem is related to https://github.com/zafarkhaja/jsemver.

RationalityFrontline:
Please use "requires static" with "org.objectweb.asm", or end users will be forced to explicitly declare an additional dependency of asm besides pf4j itself (this is mentioned in the issue #405 by @drinkitup777).

decebals:
> The remaining problem is related to https://github.com/zafarkhaja/jsemver.

For this reason I converted this PR in a draft.

decebals:
The part with jsemver remains unsolved (see https://github.com/pf4j/pf4j/pull/426#issuecomment-803242211) but I think that the direction is good.

Issue #428, opened 2021-03-08
NullPointerException when pluginid does not exist.


ahaha-vip:
org.pf4j.AbstractExtensionFinder#find(java.lang.String)

```
Set<String> classNames = findClassNames(pluginId);
// classNames may be null when pluginid does not exist.
if (classNames.isEmpty()) {
	return result;
}
```

> java.lang.NullPointerException
	at org.pf4j.AbstractExtensionFinder.find(AbstractExtensionFinder.java:177)
	at org.pf4j.DefaultExtensionFinder.find(DefaultExtensionFinder.java:65)
	at org.pf4j.AbstractPluginManager.getExtensions(AbstractPluginManager.java:591)
	at com.chengwei.jdbc.plugin.test.TestPluginManager.loadPlugin(TestPluginManager.java:44)
	at com.chengwei.jdbc.plugin.test.TestPluginManager.testDevMode(TestPluginManager.java:30)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:688)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
	at org.junit.jupiter.engine.execution.ExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.lambda$invoke$0(ExecutableInvoker.java:105)
[...]

PR #429, opened 2021-03-08
Reproduce and fix the issure #428.

ahaha-vip:


decebals:
What is wrong with this PR? I think that it's what we need. Sure, without fix, because that method should return an empty set and not NPE.

ahaha-vip:
> What is wrong with this PR? I think that it's what we need. Sure, without fix, because that method should return an empty set and not NPE.

I clicked in a wrong place, and I created a new PR #430 for this.

PR #430, opened 2021-03-08
Add a unit test in AbstractExtensionFinderTest to reproduce #428

ahaha-vip:


decebals:
@ahaha-vip 
The unit test is very good, the fix is not what I prefer. When a method return a collection (list, set, ...) is not OK nu return NULL. The  value returned in a such situation is to return an empty collection. So, please fix `org.pf4j.AbstractExtensionFinder#findClassNames`. Instead to returns NULL it will returns empty set. Thanks!

ahaha-vip:
I don't quite understand, doesn't [this fix for #428](https://github.com/pf4j/pf4j/pull/430/commits/54120929fe3128f732f11df5f358a3b37b25882b)  return an empty collection?

decebals:
> I don't quite understand, doesn't [this fix for #428](https://github.com/pf4j/pf4j/pull/430/commits/54120929fe3128f732f11df5f358a3b37b25882b) return an empty collection?

What I try to say is that you fixed the problem in https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java#L177 but the root of that problem is in https://github.com/pf4j/pf4j/blob/ff02462c9d8c208a8ddff13b3fb6e94e017a21da/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java#L221 
I want to fix the API (the public `findClassNames(String pluginId)`) and to return empty set instead of NULL.
Is it clearer now?

pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
```diff
@@ -21,13 +21,7 @@
...
 
 import java.lang.annotation.Annotation;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
decebals, line 24: In this project we don't use wildcard import. Please revert this block.

pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
```diff
@@ -218,7 +212,12 @@ public AbstractExtensionFinder(PluginManager pluginManager) {
 
     @Override
     public Set<String> findClassNames(String pluginId) {
-        return getEntries().get(pluginId);
+        Set<String> classNames = getEntries().get(pluginId);
+        if (classNames == null) {
```
decebals, line 216: Good

decebals (changes requested):

pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
```diff
@@ -174,7 +174,7 @@ public AbstractExtensionFinder(PluginManager pluginManager) {
         List<ExtensionWrapper> result = new ArrayList<>();
 
         Set<String> classNames = findClassNames(pluginId);
-        if (classNames.isEmpty()) {
+        if (classNames == null || classNames.isEmpty()) {
```
decebals, line 177: Also, please remove this modification (it's enough the initial condition `classNames.isEmpty()`).

decebals (changes requested):

pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java
ahaha-vip: https://github.com/pf4j/pf4j/blob/cf2efbd3a00219c5c43df7379bc44b28373dc298/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java#L86

And this?

decebals: > https://github.com/pf4j/pf4j/blob/cf2efbd3a00219c5c43df7379bc44b28373dc298/pf4j/src/main/java/org/pf4j/AbstractExtensionFinder.java#L86
> 
> And this?

Good catch. Yes please

decebals (changes requested):

decebals:
Thanks for your contribution!

PR #432, opened 2021-03-26
fixed remaining problems

RationalityFrontline:
The remaining problem is irrelevant to jsemver, but related to mockito and a false unit test SingletonExtensionFactoryTest.

### DIFF
--- a/pf4j/src/test/java/org/pf4j/SingletonExtensionFactoryTest.java
+++ b/pf4j/src/test/java/org/pf4j/SingletonExtensionFactoryTest.java
@@ -74,7 +74,7 @@ public class SingletonExtensionFactoryTest {
     private URL[] getClasspathReferences() throws MalformedURLException {
         String classpathProperty = System.getProperty("java.class.path");
 
-        String[] classpaths = classpathProperty.split(":");
+        String[] classpaths = classpathProperty.split(";");
         URL[] uris = new URL[classpaths.length];
 
         for (int index = 0; index < classpaths.length; index++) {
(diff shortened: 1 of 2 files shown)


pf4j/src/test/java/org/pf4j/SingletonExtensionFactoryTest.java
```diff
@@ -74,7 +74,7 @@ public void createNewEachTimeFromDifferentClassLoaders() throws Exception {
     private URL[] getClasspathReferences() throws MalformedURLException {
         String classpathProperty = System.getProperty("java.class.path");
 
-        String[] classpaths = classpathProperty.split(":");
+        String[] classpaths = classpathProperty.split(";");
```
decebals, line 77: I don't know what to say here. Probably the correct approach is to use `File.pathSeparator`.
The `;` character is used to separate the file paths on Windows.

pom.xml
```diff
@@ -51,7 +51,7 @@
 
         <junit.version>5.4.0</junit.version>
         <hamcrest.version>2.1</hamcrest.version>
-        <mockito.version>2.24.0</mockito.version>
+        <mockito.version>3.8.0</mockito.version>
```
decebals, line 54: I prefer to put this change in a separate PR.

decebals (commented):
Also, the title pf PR is very general:
```
fixed remaining problems
```
Is it related to other PR?

RationalityFrontline:
Sorry for the mess. In order to make things clear, I created two separate pr with corresponding issues: #435, #436.

PR #438, opened 2021-03-29
update module-info.java

### LINKED ISSUE #437: Problem using pf4j together with java9 module system
When using pf4j (v3.6.0) together with java9 module system, I encountered some problems:
1. Can't use slf4j of version above 1.7.25.
2. Must declare an additional dependency of asm besides pf4j itself.

I've created a demo project that could reproduce this issue: [pf4j-jpms-demo](https://github.com/RationalityFrontline/pf4j-jpms-demo/tree/master).
(To reproduce the issue, you need to edit the app/build.gradle.kts file)

RationalityFrontline:
fix #437

### DIFF
--- a/pf4j/src/main/java/module-info.java
+++ b/pf4j/src/main/java/module-info.java
@@ -26,14 +26,14 @@ module org.pf4j {
     // provides javax.annotation
     requires java.compiler;
 
-    // provided by the ASM library
-    requires org.objectweb.asm;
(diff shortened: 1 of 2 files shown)


decebals:
This PR is a duplicate of https://github.com/pf4j/pf4j/pull/426.
The big question here is what can we do with `jsemver` library. For more details see https://github.com/pf4j/pf4j/pull/426#issuecomment-803242211.

Issue #367, opened 2020-02-12
Dependency version check fails even if required is "*"


bsideup:
Hi!

During the development, I don't have a version set, and every plugin has "unspecified" as a version.
Since I do not define any contains on the version, I would expect that it will pass and won't even try to parse the provided version.

However, it seems that the parsing kicks in too early:
https://github.com/pf4j/pf4j/blob/f1dcdf8920f96f5dbb8515aeec402e9d3d7562c4/pf4j/src/main/java/org/pf4j/DefaultVersionManager.java#L42

My suggestion would be to add a check if the constraint is `*` and always return `true`.

decebals:
Can you be more specific and supply us your arguments values (`constraint` and `version`) for `DefaultVersionManager::checkVersionConstraint(String version, String constraint)`?

bsideup:
@decebals sure! Can be reproduced with:
```java
checkVersionConstraint("unspecified", "*");
```

("unspecified" comes from Gradle when the version is not set)

decebals:
@bsideup 

I will take a look. Until then you can supply your `VersionManager` custom implementation to the plugin manager via factory method [DefaultPluginManager.createVersionManager](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/DefaultPluginManager.java#L98).
Your custom version manager implementation is very simple, just extends `DefaultPluginManager.checkVersionConstraint` (put your if test in front of default implementation).

bsideup:
@decebals [already done](https://github.com/bsideup/liiklus/pull/258/files#diff-6e82f4240a2104ee2ebf32957a33810cR21), thank you for designing a very flexible API 👍

decebals:
@bsideup Can we close this issue? Do you need other information?

decebals:
Closed with #451.

PR #451, opened 2021-06-11
fixed Dependency version check fails even if required is *

wolframhaussig:
solves #367


decebals:
@wolframhaussig Thanks!

PR #452, opened 2021-06-11
add validation befor plugin can be started

### LINKED ISSUE #444: How do I avoid enabling a plugin if it will not work (eg. due to wrong OS or other external dependencies)
Hi,

What would be the correct way to not allow a plugin (or an extension in the plugin) to be enabled?

In my case, I am trying to build a plugin, which will only work when running on Linux (due to reading some /proc files).
This plugin should not be allowed to run on Windows, AIX or other OS'es where my other plugins might work fine.

I would prefer to bundle all plugins together.
I can throw an exception in the Plugin constructor, but is there a better and less violent way to do this ?

Best regards,

wolframhaussig:
closes #444

- add validate() method to plugin class
- call validate() in isPluginValid() to run the custom plugin validation
- add test for isPluginValid()


decebals:
It's too old and I see that https://github.com/pf4j/pf4j/issues/444 is closed. Please reopen it if you think it adds value.

Issue #449, opened 2021-06-04
Plugin Extensions are not detected with ClassLoadingStrategy.APD


bhanu8784:
Hi,

I am trying to use pf4j/pf4j: Plugin Framework for Java (PF4J) (github.com) with spring and wanted to use Application first strategy for class loader. 

Issue I am observing is whenever I use ClassLoadingStrategy.APD  then extensions are not identified ( I mean not able to read extensions.idx file from plugin jar). Hence pluginmanager.getExtensions(pluginid) will always return empty.

1.  My application jar (spring boot jar)  also has extensions.idx  which is always getting created  without any extensions as I don't have any extension in my application.
2.  My plugin jar also has extensions.idx with extension classes as expected.

I am setting ClassLoadingStrategy.APD for PluginClassLoader.

My Obervsation:
------------------

When extensionfinder trying to read extensions.idx it will always find one which is in my application.jar and that will not have any extensions. (getResource() method in the PluginClassloader will follow APD while loading the resource as it finds one in parent it will not attempt to load extensions from extensions.idx present in plugin jar)

Thank you
BhanuKiran

decebals:
I can confirm that is a bug but it's somehow expected. 
It's an `extensions.idx` file in app and other ``extensions.idx` in plugin. If `ClassLoadingStrategy.APD` is used, then the `extensions.idx` of app is found first.
I will take a look on the MR proposed by you.

alegacy:
Any progress on this?  I seem to have hit the same issue.   I am seeing a META-INF/extensions.idx that appears to be coming from the pf4j-spring-0.7.0.jar file which is getting used when I use the APD loading strategy.   I am not clear on why that jar file has its own extensions.idx file.

I posted a question on google groups just now:
https://groups.google.com/g/pf4j/c/52MRpvTmMOQ

``` 
2022-10-26 16:17:52.601 DEBUG 60535 --- [           main] org.pf4j.LegacyExtensionFinder           : Reading extensions storages from classpath
2022-10-26 16:17:52.602 DEBUG 60535 --- [           main] org.pf4j.LegacyExtensionFinder           : Read 'file:/Users/foo/pf4j-test-app/build/libs/pf4j-test-app.jar!/BOOT-INF/lib/pf4j-spring-0.7.0.jar!/META-INF/extensions.idx'
```

decebals:
@alegacy I didn't find a good solution yet.
[...]

PR #453, opened 2021-06-13
WIP: Plugin Extensions are not detected with ClassLoadingStrategy.APD

wolframhaussig:
solves #449 

- Added Testcase

TODO: fix Bug

The possible solutions I currently see:
1. [PluginClassLoader.getResource](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/PluginClassLoader.java#L157)
PluginClassLoader.getResource would have to be aware of the `extensions.idx` file and always use the plugin way to load the file. Downside: PluginClassLoader would have to be aware of the internals of the storage as adding a new filename would require to change the pluginClassLoader(e.g. adding a `NewExtensionStorage` which reads from `META-INF/new.idx` would require that  `META-INF/new.idx` be added to the PluginClassLoader whitelist
2. [LegacyExtensionFinder.readPluginsStorages](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java#L77)
LegacyExtensionFinder.readPluginsStorages would need to be changed from `getResourceAsStream` to `getResources`. This way, the extensionFinder would receive all possible URLs and if more than one url is returned it could decide which url to use(e.g. by comparing with plugin.getPluginPath() to get the correct url). This would partially duplicate the logic of the PluginClassLoader but would keep the classes independent.

@decebals What do you think?

### DIFF
--- a/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
+++ b/pf4j/src/main/java/org/pf4j/LegacyExtensionFinder.java
@@ -24,14 +24,21 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.net.JarURLConnection;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLConnection;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.jar.JarFile;
 
 /**
  * All extensions declared in a plugin are indexed in a file {@code META-INF/extensions.idx}.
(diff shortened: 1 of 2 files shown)


wolframhaussig:
I am not entirely happy with the solution yet but I want to get your opinion @decebals : Is this the right way to go?

decebals:
Thanks for the analysis here, and sorry for the long silence.

The bug from #449 is fixed since 3.12.1, by #586. The fix took the first option you described: `PluginClassLoader.getResource` now forces `PAD` when the requested resource is `extensions.idx`, so the plugin index wins over the one in the application. Meanwhile `LegacyExtensionFinder` became a deprecated subclass of `IndexedExtensionFinder` (#645), so this branch no longer applies to master.

I am closing this one, but two things you raised are still open and I opened issues for them:

- #673, the downside you predicted with option 1 is real. `PluginClassLoader` now hardcodes the `extensions.idx` name, so a second storage format would need another change there.
- #674, the test that should cover the parent-first case uses the parent-last class loader, so the strategy from #449 is not actually tested.

PR #454, opened 2021-06-13
Support SemverJ as Version manager

### LINKED ISSUE #440: working with maven release plugin or any plugin with verion x.y.z-LABEL
the setup:
lets say i have two plugins 
the first is plugin-a-1.0.0-SNAPSHOT.jar and the second is plugin-b-1.0.0.jar ,
now i want to make plugin-b dependent on plugin-a .
usually the most appropriate way would be to set a maven property (in plugin-b POM) called plugin-a.version and use it in both the maven dependency section and the Plugin-Dependencies manifest entry so when a version is changed you dont have to change in both places.
now the issue:
if i set plugin-a.version variable as 1.0.0-SNAPSHOT , pf4j complains that:
`com.github.zafarkhaja.semver.expr.LexerException: null
	at com.github.zafarkhaja.semver.expr.Lexer.tokenize(Lexer.java:218)
	at com.github.zafarkhaja.semver.expr.ExpressionParser.parse(ExpressionParser.java:86)
	at com.github.zafarkhaja.semver.expr.ExpressionParser.parse(ExpressionParser.java:43)
	at com.github.zafarkhaja.semver.Version.satisfies(Version.java:325)`
if i try seperating it into two variables plugin-a.version-full plugin-a.version-numeric then 1.0.0-SNAPSHOT < 1.0.0 (as expected) and the runtime files due to incompatible plugin dependencies.
since maven release plugin is in high usage - what would be the correct way to develop and check plugins with version scheme of x.y.z.-SNAPSHOT?

wolframhaussig:
closes #440

- added dependency semver4j
- added SemverJVersionManager to prevent breaking change for existing users (default implementation is still DefaultVersionManager => explicit opt-in required)
- added isStable to VersionManager interface making it easier to only allow snapshots in test environments -> just override AbstractPluginManager.isPluginValid
  - default implementation in interface returns always true to not break existing implementations
- added test method for DefaultVersionManager for snapshots
- added tests for SemverJVersionManager


decebals:
It's a good PR but for the moment I don't know what to do.
I see that the latest commit on SemverJ was on September 2020 (something in readme). I am afraid that the project will be inactive.
Probably will be a good idea to introduce SemverJ as optional dependency at runtime.

wolframhaussig:
I have updated my PR to make the library optional.

Another thought - untested yet: What would you think of using the official [maven library](https://mvnrepository.com/artifact/org.apache.maven/maven-artifact/3.8.1): the latest version is from april 2021 and we would always be compatible with the maven versioning.

From what I have found in the [javadoc](http://maven.apache.org/ref/3.8.1/maven-artifact/apidocs/index.html) you can:
- [parse versions](http://maven.apache.org/ref/3.8.1/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html#parseVersion-java.lang.String-)
- [compare versions](http://maven.apache.org/ref/3.8.1/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html#compareTo-org.apache.maven.artifact.versioning.ComparableVersion-)
- [check versions against restrictions](http://maven.apache.org/ref/3.8.1/maven-artifact/apidocs/org/apache/maven/artifact/versioning/VersionRange.html#createFromVersionSpec-java.lang.String-)

some example of maven version ranges taken from the javadoc:
- 1.0 Version 1.0
- [1.0,2.0) Versions 1.0 (included) to 2.0 (not included)
- [1.0,2.0] Versions 1.0 to 2.0 (both included)
- [1.5,) Versions 1.5 and higher
- (,1.0],[1.2,) Versions up to 1.0 (included) and 1.2 or higher

wolframhaussig:
I have added maven as alternative VersionManager - have a look and tell me what you think...

wolframhaussig:
I am not using pf4j anymore. Feel free to close this MR...

Issue #390, opened 2020-07-29
PF4J's version in fat/uber jar


MacTrophy:
When the project using pf4j is built as a fat/uber jar, pf4j detects the version from the project. I.e. lets say the project is version 1.2.3 and pf4j is 3.5.0-SNAPSHOT it will log

> PF4J version 1.2.3 in 'deployment' mode

instead of expected

> PF4J version 3.5.0-SNAPSHOT in 'deployment' mode

This is because all libraries jar files are merged in the project's jar which results in only one manifest file in the fat/uber jar which is the project's one. Honestly it's not that a big issue as long as the project's developer doesn't need to know what is pf4j's version from the logs.

Alternatives I think of could be for pf4j to use a separate file (a bit similar to the plugin.properties files) i.e. pf4j.properties to fetch the version (and other info?) from such a file instead of the manifest.

I don't know if this is considered something that is pertinent to pf4j, if not feel free to close this issue.

decebals:
I understand the problem. To be honest the problem is not critical but I really appreciate you reporting it. Thanks!
We will see. For now, I leave the issue open.

decebals:
Fixed by #455

PR #455, opened 2021-06-13
do not rely on version from Manifest

wolframhaussig:
solves #390

- added generated class Pf4jVersion to contain the current version
- added templating-maven-plugin to add project version into Pf4jVersion
- added UnitTest for version

### DIFF
--- a/pf4j/src/main/java-templates/org/pf4j/Pf4jVersion.java
+++ b/pf4j/src/main/java-templates/org/pf4j/Pf4jVersion.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2012-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pf4j;
+
+/**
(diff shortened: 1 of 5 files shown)


decebals:
It looks good. I prefer to rename `Pf4jVersion` to avoid duplication in `Pf4jVersion.VERSION` and to be able to put other useful information in the future (let's say `BUILD_NUMBER` as an example).
@wolframhaussig what do you think?

pf4j/src/test/java/org/pf4j/AbstractPluginManagerTest.java
```diff
@@ -45,5 +46,10 @@ public void getExtensionsByType() {
         List<TestExtensionPoint> extensions = pluginManager.getExtensions(TestExtensionPoint.class);
         assertEquals(1, extensions.size());
     }
+    @Test
```
decebals, line 49: please, add an empty line between methods

decebals (changes requested):

wolframhaussig:
@decebals You are right. Should I rename the class `Pf4jVersion` to `Pf4jInfo`?

decebals:
> @decebals You are right. Should I rename the class `Pf4jVersion` to `Pf4jInfo`?

I think that `Pf4jInfo` is a better name.

pf4j/src/main/java-templates/org/pf4j/Pf4jVersion.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pf4j;
+
+/**
+ * class provides access to the current Pf4j version which might otherwise not be possible(e.g. in Uber-Jars where the Manifest wass overridden)
```
decebals, line 19: I prefer to start with upper case the sentence. I try not to be annoying :smile:.

wolframhaussig:
@decebals I renamed the class and fixed the javadoc so feel free to have a look.

PR #458, opened 2021-06-14
Add code analyses via Sonar

decebals:


wolframhaussig:
@decebals I think the problem you get is that sonar is trying to validate the parent project. I have modified my build.yml run command by running 'cd pf4j' first: https://github.com/wolframhaussig/pf4j/blob/2d79fb463f5e257c88c5f48528b1f52c9074d0ec/.github/workflows/build.yml#L36

decebals:
@wolframhaussig Thanks for the assistance!

decebals:
I think that it will be nice to have an avatar icon with Sonar in the readme file (similar with build, coverage).

wolframhaussig:
I have found the following question and its response but I have not tried it myself : https://stackoverflow.com/questions/49066014/how-to-add-sonarcloud-badge-on-github

Issue #459, opened 2021-06-21
AbstractPluginManager.loadPlugins fails to load plugins and returns silently if a plugin dependency was missing


slovdahl:
Given the following scenario:
- `pluginA` and `pluginC` have no dependencies
- `pluginB` depends on `pluginA`:
   ```
   plugin-id: pluginB
   plugin-version: 1.0.0
   plugin-dependencies: pluginA
   ```
- `pluginA` is _NOT_ present in the plugin root.

When `AbstractPluginManager.loadPlugins()` is called, it returns silently and logs something like this:
 ```
2021-06-21 09:19:30 ERROR [o.p.AbstractPluginManager] :: Dependencies '[pluginB]' not found
org.pf4j.DependencyResolver$DependenciesNotFoundException: Dependencies '[pluginB]' not found
	at org.pf4j.AbstractPluginManager.resolvePlugins(AbstractPluginManager.java:794)
	at org.pf4j.AbstractPluginManager.loadPlugins(AbstractPluginManager.java:248)
```

And it also means that no other plugin is loaded (`pluginC` in this case).

Is this expected behaviour? I'm not sure what the right thing to do would be, but at least it feels like silently returning is not 100% right. Some options I could think of:
- Let the exception from the `AbstractPluginManager.resolvePlugins()` call bubble up
- Automatically detect which plugins can't be loaded and exclude them, and load the rest
- Let `AbstractPluginManager.loadPlugins()` return the plugins that could not be loaded

### Context

We call `AbstractPluginManager.loadPlugins()` during app startup, but the silent return makes it a bit challenging to detect that there was a single missing dependency, so at the moment the whole plugin system breaks down until the app is restarted if a plugin dependency was missing. We're using pf4j 3.6.0. For the time being, we can probably work around this in our own app by overriding the `loadPlugins()` and detecting the situation, but I'd be happy to submit a PR with a fix if you think it makes sense to change the behaviour.

**Comment by slovdahl** [comment]
FWIW, the same happens if there would be cyclical dependencies, the method logs and returns silently.

**Comment by slovdahl** [comment]
I now tried to override `AbstractPluginManager.loadPlugins()` and unload all plugins that have missing dependencies, but that did not work out of the box, because `AbstractPluginManager.stopPlugin(String pluginId, boolean stopDependents)` isn't able to handle `pluginState == PluginState.CREATED`. It will try to stop the plugin, but that fails with this exception:

```
java.lang.NullPointerException: null
[...]

PR #460, opened 2021-06-21
Only call `stopPlugin` during unload if the plugin is started

slovdahl:
Related to #459.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -280,6 +280,9 @@ public abstract class AbstractPluginManager implements PluginManager {
                     dependents.addAll(0, dependencyResolver.getDependents(dependent));
                 }
             }
+
+            checkPluginId(pluginId);
+
             PluginWrapper pluginWrapper = getPlugin(pluginId);
             PluginState pluginState;
             try {
@@ -466,8 +469,10 @@ public abstract class AbstractPluginManager implements PluginManager {
             return PluginState.STOPPED;
         }
 
-        // test for disabled plugin
-        if (PluginState.DISABLED == pluginState) {
+        if (PluginState.DISABLED == pluginState ||
+            PluginState.CREATED == pluginState ||
+            PluginState.RESOLVED == pluginState) {
+
             // do nothing
             return pluginState;
         }
--- a/pf4j/src/test/java/org/pf4j/DefaultPluginManagerTest.java
+++ b/pf4j/src/test/java/org/pf4j/DefaultPluginManagerTest.java
@@ -28,6 +28,8 @@ import java.nio.file.Path;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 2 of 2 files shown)


slovdahl:
Unrelated to this PR, but I just noticed it at the same time.

```java
public abstract class AbstractPluginManager implements PluginManager {

    // ...

    protected boolean unloadPlugin(String pluginId, boolean unloadDependents) {
        try {
            if (unloadDependents) {
                List<String> dependents = dependencyResolver.getDependents(pluginId);
                while (!dependents.isEmpty()) {
                    String dependent = dependents.remove(0);
                    unloadPlugin(dependent, false);
                    dependents.addAll(0, dependencyResolver.getDependents(dependent));
                }
            }
        }

        // ...
    }
}
```

Is there a reason `unloadPlugin(dependent, false)` is called with `false` as the second parameter? To me, it would somehow feel more natural to use `true`, to make sure that the whole chain of dependant plugins are unloaded first.

decebals:
@slovdahl 
I simply forgot to answer this issue. Sorry.
I think that we need some tests here.
In my opinion the problem is in [AbstractPluginManager#stopPlugin](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L451). We need to stop a plugin only if it is in STARTED state. Method AbstractPluginManager#stopPlugin should check the state.

> Is there a reason unloadPlugin(dependent, false) is called with false as the second parameter? To me, it would somehow feel more natural to use true, to make sure that the whole chain of dependant plugins are unloaded first.

It's a good question and we need to check the impact of this change.

Also, we need to add more tests for AbstractPluginManager, to cover more cases (transitions from one state to another).

perlun:
@slovdahl - should we also try to include this change? In `unloadPlugin()`, right before the`firePluginStateEvent()` call (could unfortunately not make it a suggestion here in the MR because of GitHub limitations): 

```java
// Differs from the pf4j 3.6.0 superclass: clean up the unresolved plugins state,
// otherwise loading the same plugin again might not succeed.
getUnresolvedPlugins().remove( pluginWrapper );
```

slovdahl:
> @slovdahl - should we also try to include this change from our version of the class? In `unloadPlugin()`, right before the`firePluginStateEvent()` call (could unfortunately not make it a suggestion here in the MR because of GitHub limitations):
> 
> ```java
> // Differs from the pf4j 3.6.0 superclass: clean up the unresolved plugins state,
> // otherwise loading the same plugin again might not succeed.
> getUnresolvedPlugins().remove( pluginWrapper );
> ```

It's probably out of scope in this PR, I wrote a test in #558 and aim for fixing the relevant problems there. I think this is one of them.

slovdahl:
Also looks like #521 by @decebals is a much more complete fix than my suggestion here.

decebals:
@slovdahl I tested your test from this PR with #521 and it passes.

slovdahl:
@decebals Nice :+1: Feel free to pull in the test from this PR in #521, and we can close this.

decebals:
@perlun Can you test the fix (available on master already and available as 3.11.0-SNAPSHOT)?

My intention is to release a new version after #564 will be merged.

perlun:
Sounds good @decebals, I'll make sure that either me or @slovdahl tests the 3.11.0-SNAPSHOT version. Does it include the https://github.com/pf4j/pf4j/pull/567 fix too? If so, that would be super-great. :bow:

decebals:
I [uploaded](https://oss.sonatype.org/content/repositories/snapshots/org/pf4j/pf4j/3.11.0-SNAPSHOT/) the latest SNAPSHOT version (it contains what it's on master - included #567).
If your tests go well, I can release a new version.

decebals:
> I [uploaded](https://oss.sonatype.org/content/repositories/snapshots/org/pf4j/pf4j/3.11.0-SNAPSHOT/) the latest SNAPSHOT version (it contains what it's on master - included #567).
> If your tests go well, I can release a new version.

@slovdahl @perlun How are your tests going? I need a good result to make the release.
​

slovdahl:
@decebals I'm sorry for keeping you waiting here, I haven't been able to prioritize this the past weeks. Today I finally have a bit of time.

I'm doing some testing as we speak, but I found some issues on our side that I need to investigate and fix before I can test all our use-cases. But it looks like we can remove all of our duplication of the `AbstractPluginManager` methods and use `PluginStateListener` instead :+1: :partying_face:

slovdahl:
Finally got the issues sorted out. Everything seems to work great, so as far as I'm concerned, feel free to put out a new release :slightly_smiling_face: Thanks a lot for the all work on this, greatly appreciated! :bow:

PR #464, opened 2021-07-20
Refactoring annotation processor

centoria:
Developer notes :
I've debugged the way how the ExtensionAnnotationProcessor works and found many problems.
1- the bug in the LegacyExtensionStorage, does not handle some IO exceptions,
that's why the .idx is generated once (the first round of the annotation process), so when the newly generated files with annotation @Extension , it gets detected, but the .idx is not changed
2- the Map oldExtensions, is dead code and does not work  (because the file is modified once) + if we always generating the file, we don't need to get oldExtensions from the .idx file, because in the new rounds we got all extensions any way (modifié) 

CEO / CTO of Centoria remarqs :

We absolutely need this fix, because we need to generate classes with @Extension and be compiled in the second round.

Thks a lot.


decebals:
@centoria 
It's some noise (formatting, increase version) in this PR and from this reason it's difficult for me to follow it.
First of all each PR should reference an issue to have some context. I was not aware of any problems on this subject (APT) and I am not sure from what I see in this PR that they are some issue.
I don't understand why this PR increments version, I don't think that is OK.
Some resources (reader, ..) are close from outside and from this reason in annotation processor in some places (I see a single place until now) the resource is not closed at the end of the method.
The `oldExtensions` are used in `ServiceProviderExtensionStorage`. From what I remember it's about write/import extensions found in `extensions.idx` (pf4j default storage) in the Java `ServiceLoader`/`ServiceProvider` storage generated/updated  automatically by pf4j.
I appreciate very much new more unit tests in `ExtensionAnnotationProcessorTest` that should reproduce the issues found by you.

decebals:
It's too old and we have a lot of conflicts.

PR #478, opened 2021-12-12
Upgrade to Log4j version to 2.15.0

zrashwani:
Updating log4j library version to 2.15.0 to handle [**CVE-2021-44228**](http://cve.mitre.org/cgi-bin/cvename.cgi?name=2021-44228) to patch remote code execution vulnerability in Apache log4j


decebals:
Thanks!

Just as a clarification, the `log4j` library is used only in `maven-archetypes/quickstart` and `demo`.

PR #485, opened 2022-01-19
Update README url

solonovamax:
Update README url to avoid SSL_ERROR_BAD_CERT_DOMAIN error when connecting to side

Currently, when you click on the link in the README, you get this very nice page (firefox):

image

This change fixes that by updating `http://www.pf4j.org` -> `http://pf4j.org`


decebals:
I checked and the problem appears only on Firefox browser. From what I saw the problem is more complicated.
In the end it's a good quick fix, thanks!

PR #450, opened 2021-06-07
add wrapper to plugin manager

wolframhaussig:
solves #353 

- extracted PluginWrapper creation into separate method `createPluginWrapper`
- added `SecurePluginManagerWrapper` for users who want to prevent plugins having full access to pluginManager which can be used like this:
```java
public class SecurePluginManager extends DefaultPluginManager  {
    @Override
    protected PluginWrapper createPluginWrapper(PluginDescriptor pluginDescriptor, Path pluginPath, ClassLoader pluginClassLoader) {
    	PluginWrapper pluginWrapper = new PluginWrapper(new SecurePluginManagerWrapper(this, pluginDescriptor.getPluginId()), pluginDescriptor, pluginPath, pluginClassLoader);
        pluginWrapper.setPluginFactory(getPluginFactory());
        return pluginWrapper;
    }
}
```
- added tests for `SecurePluginManagerWrapper`

Open points:
- ~Test for `whichPlugin` fails because somehow, `pluginManager.getExtensionClasses(THIS_PLUGIN_ID)` seems to load the extension `TestExtension` from the application classloader instead of the plugin classloader.~ was caused by the PLUGIN_PACKAGE_PREFIX check in `PluginClassLoader`

Issue #372, opened 2020-04-21
PF4J 4 roadmap


decebals:
I think that is time to talk about PF4J 4.x roadmap.
PF4J is maintained under the [Semantic Versioning](https://semver.org) guidelines, so in a new major version (4.0 in our case) we can make incompatible changes in our API.
Now is time to improve PF4J without to care about backward compatible.

I will try to add some proposes:

- [ ] Migrate to Java 11
- [ ] Async APIs (#269)
- [ ] List all extension points and extensions for a plugin (#207)

The list with proposes will grow in time.

If you think that something is bad designed or you want to change something in the current API, please add a comment.

rmrodrigues:
Hi @decebals ,
I think we have to create a more complex DEMO app.
Example: Create a web-app to export a list of data where each plugin represents export formats such as:

MS Office plugin:

- Excel
-CSV
- Word
Open Office plugin:

- ODT

Generic plugin:

- Text

This dame could provide a plugin manager web based with the following actions:

- Install/uninstall plugins
- Start/Stop plugins

decebals:
Hi @rmrodrigues 
Now, the PF4J ecosystem is big enough. I see a lot of application that uses PF4J with success with Vert.x, Spring Boot. I think we don't lack projects that show how to use PF4J.
I want to focus on PF4J and not on how to integrate Excel, Word or something similar.

rmrodrigues:
@decebals  the idea is not to integrate with Excel and Word. It's just a more complex use case using the framework.

fendo8888:
Looking forward to an example

decebals:
@fendo8888 What is wrong with [demo](https://github.com/pf4j/pf4j/tree/master/demo) (https://pf4j.org/doc/demo.html)?

fendo8888:
@decebals I have seen that demo, I mean an example of the integrity of a replication point

fendo8888:
When will version 4.0 come out, big brother

decebals:
@fendo8888 Nothing concrete until now, I cannot give you an estimation. It's a community driven project so the roadmap is entirely dictated my community. As I have not seen a high demand for the points proposed in this issue, so it is not a top priority :).
We will see.

hank-cp:
[...]

PR #512, opened 2023-01-15
Relax Plugin construction (remove dependency on PluginWrapper)

decebals:
The idea is to remove `PluginWrapper` from `Plugin` and add possibility to inject an application base `PluginContext` using a custom `PluginFactory`.
In few words instead of:
```java
public class WelcomePlugin extends Plugin {

    public WelcomePlugin(PluginWrapper wrapper) {
        super(wrapper);
    }

}
```
now we have:
```java
public class WelcomePlugin extends DemoPlugin {

    public WelcomePlugin(PluginContext context) {
        super(context);
    }

}
```

where `PluginContext` is a custom class that facilitate the communication of the plugin with the external ecosystem (application , plugin manager, ..) and whose structure differs from application to application.

Also, this PR add support for default (no parameters) constructor in `Plugin`-> in many situation we don't need extra information (a plugin context).

I deprecated the part (constructor, field, getter) with `PluginWrapper` from `Plugin`. I think for now this is the best solution because they are many applications with many plugins based on PF4J and we cannot broke the API (remove `PluginWrapper` from `Plugin`)

Also, see https://github.com/pf4j/pf4j/pull/450#issuecomment-859277855.

### DIFF
--- a/demo/api/src/main/java/org/pf4j/demo/api/DemoPlugin.java
+++ b/demo/api/src/main/java/org/pf4j/demo/api/DemoPlugin.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -13,20 +13,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.pf4j.test;
+package org.pf4j.demo.api;
 
 import org.pf4j.Plugin;
 
 /**
- * A wrong {@link org.pf4j.Plugin}.
- * It's wrong because it calls super constructor with {@code null} for ({@link org.pf4j.PluginWrapper} parameter).
+ * Base {@link Plugin} for all demo plugins.
  *
- * @author Mario Franco
+ * @author Decebal Suiu
  */
-public class AnotherFailTestPlugin extends Plugin {
+public abstract class DemoPlugin extends Plugin {
 
-    public AnotherFailTestPlugin() {
-        super(null);
+    protected final PluginContext context;
+
+    protected DemoPlugin(PluginContext context) {
+        super();
+
+        this.context = context;
     }
 
 }
--- a/demo/api/src/main/java/org/pf4j/demo/api/PluginContext.java
+++ b/demo/api/src/main/java/org/pf4j/demo/api/PluginContext.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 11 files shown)


decebals:
As you may have noticed, the `PluginContext` that is injected in `Plugin` via constructor is a custom class (defined in demo application) and it's not part of PF4J.
Just as well, instead of `PluginContext`, you can inject a simple `Map<String, Object>` as plugin context.
If you have nothing to send to Plugin, then do nothing. In this case is up to you if in Plugin you declare a no parameters constructor or not.

I opted for removing `PluginWrapper` from `Plugin` because `PluginWrapper` is somehow an internal component. 
The version with `PluginWrapper` is somewhat forced, initially it was to pass some runtime context to `Plugin`, but along the way it became clear that this information related to context differs from application to application and it's not good to force a contract here.

I think that the best solution to add this change is to deprecate the things (field, constructor, getter) in Plugin class and to remove the deprecated code in next major code. Another variant is to remove the code related PluginWrapper from Plugin in next PF4J major version. If you have other better ideas, please let me know.
A code review is welcome, too.

perlun:
> I think that the best solution to add this change is to deprecate the things (field, constructor, getter) in Plugin class and to remove the deprecated code in next major code.

I agree, it's good to see that this wasn't removed in 3.9.0 but only deprecated. Removing it would have been unexpected towards the consumers of the library.

Perhaps worth maintaining a `RELEASE-NOTES.md` file in the repo or something, where this could be planned/noted down already? That way, we would have a good place to store "things that we want to change for 4.0", both for remembering it ourselves but also for people to read and be prepared for what's coming. (Obviously, the `@Deprecation` also helps with the latter)

decebals:
We have am issue related to PF4J 4 roadmap. It's about https://github.com/pf4j/pf4j/issues/372.
The idea with deletion is that if this happens in the core of the application it's not a problem, I would say, but if the change happens through plugins this is a problem because you cannot force an upgrade in all plugins.

perlun:
>  The idea with deletion is that if this happens in the core of the application it's not a problem, I would say, but if the change happens through plugins this is a problem because you cannot force an upgrade in all plugins.

Valid point yes. :+1: I personally think the world would be a slightly better place if everyone took a quite strict adherence to [SemVer](https://semver.org/), which I am sure you are aware of already. The basics, for those of us who might not have worked with it so much:

* Add new functionality in a backwards-compatible way in minor releases (3.10.0, 3.11.0, 3.12.0 and so forth).
* Add bugfixes in patch-level releases (3.10.1, 3.10.2, etc)
* Make breaking changes (remove, modify parts) in the public API **only** in major releases (3.0, 4.0, 5.0 and so forth).

Then the question is of course "what is defined as a public API", but in this case I'd say it's anything that would cause a compilation error (be it in the application or in any of the plugins). All such changes are fine, but they should (with a SemVer world view) _only_ ever be done in a major version upgrade. That makes the upgrade path predictable for everyone: you can always safely upgrade to a minor release or a patch-level release, you (almost) never have to make any changes in your application apart from bumping the dependency version. Major releases can be a different story but this is what is communicated by increasing the major version.

decebals:
PF4J uses semantic versioning.
The idea is that we do not have (yet) a significant number of changes (new features, API changes) to justify a new major version (4).
On the other side, I prefer to maintain a single major version. It is difficult for me to do maintenance on several major versions, to port certain fixes, improvements on older versions.
​

perlun:
Yeah, I know strict adherence to Semantic Versioning can be challenging. :thinking: Thanks for doing your best. :+1:

Issue #491, opened 2022-03-19
Fix modules for java 17 JPMS.


MXP2095onetechguy:
Experiencing `java.lang.module.FindException: Module slf4j.api not found, required by org.pf4j` when using maven with a modular project. Can this problem be fixed?

Sailsman63:
This is not a problem so much as an outdated dependency. PF4J currently pulls in slf4j 1.7.30, which is not modularized.

Potential workarounds:

- Upgrade the `slf4j-api.jar` that you use to 1.8.0-beta4. (or the 2.0.0 alpha builds - they are still back-compatible from a client perspective. Just make sure that you use a matching implementation for your actual logging) This should be doable with "shading" in your build system.
- Accept that this version of slf4j is unmodularized, so needs to be loaded as part of the un-named module on the *classpath* (as opposed to the module-path)
- Auto-module shenanigans
- build PF4J yourself, with the slf4j version upgraded

decebals:
See https://github.com/pf4j/pf4j/issues/411

Sailsman63:
True, and a perfectly valid decision for this project to make.

On the user end, though, it's worth noting that the 2.0.0Alpha versions are in sort of a meta-stable state: They still support all of the legacy API, and have been in successful production use by large projects, such as Jetty, for some time. From an end-user perspective, swapping them in for deployment might well be worthwhile.

decebals:
So, for the moment I will close this issue (some workarounds are proposed by @Sailsman63 and we don't have a GA version of slf4j that are JPMS ready).
Please reopen it if you have other questions/ambiguities.

decebals:
I see that slf4j 2.x is available in Maven central repository. For me is

Sailsman63:
Log4J 1.x does have some security issues. The slf4j author (who also wrote the original log4j code) has released a hardened update for those who really don't want to swap out their log configurations. It's called reload4j, just in case anyone is wondering.

decebals:
@MXP2095onetechguy Can you try with the latest version of PF4J (in the context of https://github.com/pf4j/pf4j/pull/514)?

decebals:
Fixed by https://github.com/pf4j/pf4j/pull/514.

Issue #411, opened 2020-11-23
please update the version of slf4j


l1046j:
org.slf4j.ext.EventData in the slf4j-ext module in QOS.CH SLF4J before 1.8.0-beta2 allows remote attackers to bypass intended access restrictions via crafted data.

decebals:
Can you submit a PR please?

decebals:
We have https://github.com/pf4j/pf4j/pull/412 for this but for the moment I closed because it is debatable if it is OK to add a beta version as a dependency in the project

PR #514, opened 2023-02-01
Switch to slf4j-2.x and replace log4j with slf4j-simple

decebals:
Solve https://github.com/pf4j/pf4j/issues/491 and other issues (https://github.com/pf4j/pf4j/issues/411, ..) related to log4j vulnerabilities.


decebals:
Does anyone see any problems with this modification?

Issue #536, opened 2023-08-12
The method of extracting the zip file has a path traversal vulnerability


afeng2016-s:
## description
Dear project developers, I use SpringBoot and pf4j to implement the system's extension plug-in function, the use of zip or jar package format is very easy to expand the system. When I was using pf4j, I found that the pluginPath value was not verified securely when loading plug-ins. If the developers using pf4j were not security conscious, directly calling the loadPluginFromPath method to receive malicious parameters passed by untrusted users would lead to directory traversal vulnerabilities.

## affected version
<= release-3.9.0

## [vulnerability analysis](javascript:;)
The sample code attempts to extract a malicious file (for example: C:\Windows\notepad.exe) to the root path of drive E to simulate a directory traversal attack.
1. Build a malicious ZIP file，Save the resulting ZIP file to E:\\Code\\0811\\malicious.zip。
![image](https://github.com/pf4j/pf4j/assets/85676107/44f28efb-dd32-4695-b646-5c35f1df74d6)

2, write a test class, call DefaultPluginManager#loadPluginFromPath method to load E:\\Code\\0811\\malicious.zip
![image](https://github.com/pf4j/pf4j/assets/85676107/76b4e717-cf2d-4773-9a86-b3be5bc55972)

First run a test class to prove that the path E:\Windows\notepad.exe does not exist.
![image](https://github.com/pf4j/pf4j/assets/85676107/f895e117-c6c7-492e-bdfa-194bff0352ee)

3. Debug the code
Go to DefaultPluginManager#loadPluginFromPath.
Load the plug-in from disk. If the path is a zip file, unzip it first.
![image](https://github.com/pf4j/pf4j/assets/85676107/46477780-15bb-479a-9c71-c1ddb0fc2427)

Do some preparatory work in FileUtils
![image](https://github.com/pf4j/pf4j/assets/85676107/fa1bc2b1-4392-4553-be5d-7d4d67cd6ec9)

Follow up unzip.extract()
![image](https://github.com/pf4j/pf4j/assets/85676107/d2115c93-8fef-4620-b978-57f9032a5dd6)

![image](https://github.com/pf4j/pf4j/assets/85676107/c9a14ad1-851a-4387-9c7e-a573afcc134f)

After the execution is complete, extract the target file to the root directory of drive E.
![image](https://github.com/pf4j/pf4j/assets/85676107/ea3352a6-91d4-4283-9b2a-23a641ce876a)

[...]

PR #537, opened 2023-08-13
Add security checks to prevent directory traversal when decompressing…

afeng2016-s:
## This is a PR submission for #536

To verify that there is a directory traversal risk when unzipping the zip file, I test in FileUtilsTest.java.

1.Using the zipslip vulnerability, create a zip file.Save the created zip file in the D:/code/pf4j directory, if you do not have this path on your computer D drive, create it.
image

2. Next, call expandIfZip or loadPluginFromPath method to extract the zip file to the root directory of disk D of the computer.
image

3.To prevent path crossing problems caused by unsafe input, I recommend adding checks to the extract() method.
image

After adding the check, an exception is thrown when there is a malicious file name
image

Sorry, commits/c1b03c92c03cc42ef7d197d962acd785bbea60dd is wrong, commits/ed9392069fe14c6c30d9f876710e5ad40f7ea8c1 provide repair plan is correct.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/util/FileUtils.java
+++ b/pf4j/src/main/java/org/pf4j/util/FileUtils.java
@@ -185,11 +185,6 @@ public final class FileUtils {
         String directoryName = fileName.substring(0, fileName.lastIndexOf("."));
         Path pluginDirectory = filePath.resolveSibling(directoryName);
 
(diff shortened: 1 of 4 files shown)


decebals:
Replaced by https://github.com/pf4j/pf4j/pull/538.
@afeng2016-s Thank you for your implication!

PR #545, opened 2023-09-14
unload broken plugins fix

finn0s:
The problem is described in #546


decebals:
@finn0s 
It look good, thank you!
Before I merge this PR, please remove the `final` word from changes.

PR #551, opened 2023-10-04
Write empty extensions descriptor even if no extensions are found

### LINKED ISSUE #550: LegacyExtensionFinder finds duplicate extensions
Setup: the `ClassLoader` of a plugin can access classes - and resources of itself and also of all plugins it depends on. `LegacyExtensionFinder` is used to read information about available extensions from `extensions.idx`.

Scenario:
 - plugin A provides some extensions
 - plugin B provides extension points but no extensions on its own. It also depends on plugin A. Since it doesn't have any extensions, there is no `extensions.idx` bundled.
- when asking the classloader of plugin B for the `extensions.idx` resource as stream, it falls back to the one bundled with plugin A since that is on the classpath, too, it being a dependency. Consequently, extensions are registered twice.

Expected:
  - no extensions for plugin B are found, extensions of plugin A are found once

Possible fixes:
  - write an empty `extensions.idx`, even if no extensions are found (quite straightforward)
  - replace `getResourceAsStream`  with `getResource` and check that the URL isn't processed twice. Unfortunately prone to error as the extensions could still show up as coming from the wrong plugin. Also ugly because of shared mutable state.
  - ?

milgner:
Closes #550 

I tried this only with the legacy mechanism. Not 100% sure whether it works with JDK9 module mechanism, too - but I see no reason that it shouldn't.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
+++ b/pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
@@ -130,10 +130,10 @@ public class ExtensionAnnotationProcessor extends AbstractProcessor {
             }
         }
 
-        // write extensions
-        if (extensions.size() > 0) {
-            storage.write(extensions);
-        }
+        // Even an empty extensions descriptor is semantically correct and should be
+        // written to prevent classloaders from falling back to extension descriptor
+        // resources from dependencies of the plugin being processed.
+        storage.write(extensions);
 
         return false;
     }


decebals:
> CI complains about a missing Sonar token - is there anything I can do about it? 🤔

It is a problem known to me and for the moment I don't have a quick fix. I will continue to investigate.

pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
```diff
@@ -130,10 +130,10 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
             }
         }
 
-        // write extensions
-        if (extensions.size() > 0) {
-            storage.write(extensions);
-        }
+        // even invoke if no extensions were found: an empty extensions descriptor
+        // is semantically correct and prevents classloaders falling back to descriptors
```
decebals, line 134: > prevents classloaders falling back to descriptors from dependendencies

Can you add more details?

decebals: I see now that the problem is described in https://github.com/pf4j/pf4j/issues/550.

milgner: Indeed! I checked the patch in the context of my project and the extensions are looked up correctly again.

pf4j/src/main/java/org/pf4j/processor/ExtensionAnnotationProcessor.java
```diff
@@ -130,10 +130,10 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
+        // from dependendencies
```
decebals, line 135: Please fix the typo in `dependendencies`.

milgner: Done with a bit of rewording to (hopefully) improve documentation of what's going on.

PR #558, opened 2024-01-22
Fix potential corrupt state in AbstractPluginManager

slovdahl:
Related to #459.

### DIFF
--- a/pf4j/src/test/java/org/pf4j/DefaultPluginManagerTest.java
+++ b/pf4j/src/test/java/org/pf4j/DefaultPluginManagerTest.java
@@ -190,4 +190,44 @@ public class DefaultPluginManagerTest {
         assertFalse(pluginJar.file().exists());
     }
 
+    @Test
+    public void loadingPluginWithMissingDependencyDoesNotBreakOtherPlugins() throws IOException {
+        // Load 2 plugins, one with a dependency that is missing and one without any dependencies.
+        PluginZip pluginZip1 = new PluginZip.Builder(pluginsPath.resolve("my-first-plugin-1.1.1.zip"), "myPlugin1")
+            .pluginVersion("1.1.1")
+            .pluginDependencies("myBasePlugin")
+            .build();
+
+        PluginZip pluginZip2 = new PluginZip.Builder(pluginsPath.resolve("my-second-plugin-2.2.2.zip"), "myPlugin2")
+            .pluginVersion("2.2.2")
+            .build();
+
+        pluginManager.loadPlugins();
+        pluginManager.startPlugins();
+
+        // myPlugin2 should have been started at this point.
+        assertEquals(PluginState.STARTED, pluginManager.getPlugin(pluginZip2.pluginId()).getPluginState());
+
+        pluginManager.unloadPlugin(pluginZip1.pluginId());
+
+        // No traces should remain of myPlugin1.
+        assertTrue(
+            pluginManager.getUnresolvedPlugins().stream()
+                .noneMatch(pluginWrapper -> pluginWrapper.getPluginId().equals(pluginZip1.pluginId()))
+        );
+
+        // Load the missing dependency, everything should start working.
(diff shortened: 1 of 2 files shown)


pf4j/src/test/java/org/pf4j/DefaultPluginManagerTest.java
```diff
@@ -190,4 +190,44 @@ public void deleteJarPlugin() throws Exception {
...
+
+        PluginZip pluginZip2 = new PluginZip.Builder(pluginsPath.resolve("my-second-plugin-2.2.2.zip"), "myPlugin2")
+            .pluginVersion("2.2.2")
+            .build();
+
+        pluginManager.loadPlugins();
+        pluginManager.startPlugins();
+
+        // myPlugin2 should have been started at this point.
+        assertEquals(PluginState.STARTED, pluginManager.getPlugin(pluginZip2.pluginId()).getPluginState());
```
slovdahl, line 209: Right now, the test will fail at this point.

slovdahl: Some suggested fixes pushed in https://github.com/pf4j/pf4j/pull/558/commits/b96bb93b8ceb44a43c6bb95df3057bbd59c471f6.

decebals:
I test the test supplied by you in this PR, it passes but the test introduced in #460 failed at https://github.com/pf4j/pf4j/pull/460/files#diff-6dbe4409bbc76e0eb975d634fb3aa8deac4debe8e9cdbb872812861f5b304056R206

decebals:
From what I understand this part with `unloadPlugin` is sensible because the plugin is removed from list with plugins.
I think that it's fine to have all plugins in list but with different states.

decebals:
The current `AbstractPluginManager#resolvePlugins`  comes with an ALL OR NOTHING implementation . According [code](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L791) if they plugins with missing dependencies the entire resolve plugins process fails (an exception is thrown) . I must admit that in this scenario to suppress the exception in `loadPlugins` (only an error message in log) is not perfect because the application does not have a feedback that the process of loading plugins went as it should.

decebals:
On the other side to go only with the plugins that could be solved, I don't think it's a fit all solution. Maybe we need to implement a strategy pattern here and the application to choose one.

slovdahl:
> I test the test supplied by you in this PR, it passes but the test introduced in https://github.com/pf4j/pf4j/pull/460 failed at https://github.com/pf4j/pf4j/pull/460/files#diff-6dbe4409bbc76e0eb975d634fb3aa8deac4debe8e9cdbb872812861f5b304056R206

Ah, yes, of course :smile: Because this change avoids the problem in the first place (that we need to manually unload a plugin that was partially loaded).

slovdahl:
> From what I understand this part with unloadPlugin is sensible because the plugin is removed from the list with plugins.
I think that it's fine to have all plugins in list but with different states.

So you think the two changes in `AbstractPluginManager#unloadPlugin` could be merged as is? In that case I can extract them to a separate PR, and see if I can write some specific tests for them.

slovdahl:
> I must admit that in this scenario to suppress the exception in loadPlugins (only an error message in log) is not perfect because the application does not have a feedback that the process of loading plugins went as it should.

:+1: 

> On the other side to go only with the plugins that could be solved, I don't think it's a fit all solution. Maybe we need to implement a strategy pattern here and the application to choose one.

Yep, agreed, this might not be for everyone. In our context, it's what we want, but some kind of strategy sounds like an even better solution. Do you have any concrete ideas in mind?

decebals:
@slovdahl 
> Do you have any concrete ideas in mind?

Yes, see https://github.com/pf4j/pf4j/pull/564.

decebals:
Replaced by https://github.com/pf4j/pf4j/pull/564.

PR #560, opened 2024-01-29
Upgrade Java SemVer to 0.10.2

JohnNiang:
Recently, I have noticed that the project <https://github.com/zafarkhaja/jsemver> has become active again and they have releases some new versions. See <https://github.com/zafarkhaja/jsemver/releases> for more.

This PR upgrades the dependency to [0.10.2(latest)](https://github.com/zafarkhaja/jsemver/releases/tag/v0.10.2).


decebals:
Thanks!

PR #564, opened 2024-02-05
Add strategy for handling the recovery of a plugin that could not be resolved

decebals:
Strategy for handling the recovery of a plugin that could not be resolved (loaded) due to a dependency problem.
Possible values:

- THROW_EXCEPTION -> Throw an exception when a resolve (load) failure occurs
- IGNORE_PLUGIN_AND_CONTINUE -> Ignore the plugin with the resolve (load) failure and continue;
the plugin with problems will be removed/unloaded from the plugins list

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -109,6 +109,7 @@ public abstract class AbstractPluginManager implements PluginManager {
     protected boolean exactVersionAllowed = false;
 
     protected VersionManager versionManager;
+    protected ResolveRecoveryStrategy resolveRecoveryStrategy;
 
     /**
      * The plugins roots are supplied as comma-separated list by {@code System.getProperty("pf4j.pluginsDir", "plugins")}.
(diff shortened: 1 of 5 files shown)


decebals:
@slovdahl 
Can you test it, in context https://github.com/pf4j/pf4j/pull/558 and https://github.com/pf4j/pf4j/issues/459?
All tests added by you are present in this PR.

slovdahl:
Thanks @decebals, and sorry for the delay. I'll definitely try this out, if not today, early next week.

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -1058,4 +1038,88 @@ protected <T> List<T> getExtensions(List<ExtensionWrapper<T>> extensionsWrapper)
...
+        return resolveDependencies();
+    }
+
+    /**
+     * Retrieve the strategy for handling the recovery of a plugin resolve (load) failure.
+     * Default is {@link ResolveRecoveryStrategy#THROW_EXCEPTION}.
+     *
+     * @return the strategy
+     */
+    protected ResolveRecoveryStrategy getResolveRecoveryStrategy() {
```
slovdahl, line 1095: Should this method be `final`? Right now, it's unclear what overriding it would mean. It's not used by `AbstractPluginManager` itself, so one cannot override it as a means for setting it.

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -1058,4 +1038,88 @@ protected <T> List<T> getExtensions(List<ExtensionWrapper<T>> extensionsWrapper)
...
+     */
+    protected void setResolveRecoveryStrategy(ResolveRecoveryStrategy resolveRecoveryStrategy) {
+        this.resolveRecoveryStrategy = resolveRecoveryStrategy;
+    }
+
+    /**
+     * Strategy for handling the recovery of a plugin that could not be resolved
+     * (loaded) due to a dependency problem.
+     */
+    enum ResolveRecoveryStrategy {
```
slovdahl, line 1112: I think this needs to be at least `protected`, otherwise it cannot be set by subclasses:

```suggestion
    protected enum ResolveRecoveryStrategy {
```

slovdahl (approved):
This looks like a great way to solve the issue 👍 Thank you so much for this effort.

I built a local SNAPSHOT version of this PR and updated our application to use it. We have some tests for the workarounds we have had in place in our `DefaultPluginManager` subclass, and after removing all the workarounds the tests still pass. So far, so good 🙂 I will still to do a bit more testing of it, but overall I think this is good to go.

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -280,56 +281,54 @@ public boolean unloadPlugin(String pluginId) {
...
-                ClassLoader classLoader = pluginClassLoaders.remove(pluginId);
-                if (classLoader instanceof Closeable) {
-                    try {
-                        ((Closeable) classLoader).close();
-                    } catch (IOException e) {
-                        throw new PluginRuntimeException(e, "Cannot close classloader");
-                    }
+            log.info("Unload plugin '{}'", getPluginLabel(pluginWrapper.getDescriptor()));
+        } catch (Exception e) {
+            pluginState = PluginState.FAILED;
```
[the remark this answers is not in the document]
decebals: Yes, we can

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
decebals: Done

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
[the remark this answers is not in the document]
decebals: Done

pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
decebals: Done

decebals:
> and after removing all the workarounds the tests still pass

If you have more tests that can be added directly to pf4j, I would be happy to add them.

slovdahl:
> If you have more tests that can be added directly to pf4j, I would be happy to add them.

Good idea, I'll see if we have some tests that are easy to adapt to pf4j and make sense in that context.

decebals:
Please tell me if I can merge this PR in its current form or if you see any problems and this PR needs to be updated.

slovdahl:
Did a bit more testing and AFAICT it just works :tm: So feel free to merge it in the current form. Thanks once again for the effort! :bow:

aschwarte10:
@decebals, @slovdahl  is there a technical reason why you made `getResolveRecoveryStrategy` a final method?

We tried to update to the 3.11 release and for us it fails because of dependency injection: we have a bean extending  `DefaultPluginManager` and this can no longer be injected due to the `final`.

If there is no technical reason, would it be an option to remove the final in a next release? If so, I would like to creat this as issue.

slovdahl:
https://github.com/pf4j/pf4j/pull/564#discussion_r1486593654

In general I think it makes a lot of sense to mark methods that cannot reasonably be overridden as `final`.

What kind of environment are you using, Spring version etc? We're also extending `DefaultPluginManager` and it's also a Spring bean, but it works fine in our context. Spring 5.x, Java 17.

aschwarte10:
Thanks for the fast response @slovdahl 

We are using Jakarta CDI and JBoss Weld as an implementation (in a Java 17 environment). This framework makes use of proxies for the implementation, which do not support final modifiers in certain places. See https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0#unproxyable

Currently, this "final" modifier on the protected method prevents us from upgrading pf4j. I agree that it can make. sense to have final modifiers, but here it prevents us from using the DefaultManager as-is.

decebals:
I don't see a problem to remove the `final` word from the method signature. 
In the end it is important to have happy users :smile:.

aschwarte10:
@decebals sounds great!

Then I propose to track all information in a quick issue (I will do this) and you can address according to the project needs. For us it would be great if you can communicate some kind of timeline such that we we can internally align our planning

Thanks again 👍

decebals:
Please create an issue. We can do a quick fix release

PR #567, opened 2024-02-19
Post `PluginState.UNLOADED` event when plugins are unloaded

### LINKED ISSUE #198: Adding a new PluginState "UNLOADED" may be useful.
I have a questions about the [Plugin lifecycle](http://www.pf4j.org/doc/plugin-lifecycle.html)
When can the plugin be removed ? I mean the plugin Jar-file or folder can be removed by system, like by the command ```rm``` or ```del```, not only by the code.

In [AbstractPluginManager.java](https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L299), there is a method , ```deletePlugin```, 
It requires the plugin **unload** by the method [unloadPlugin](https://github.com/decebals/pf4j/blob/master/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L248) first. 
```java
    @Override
    public boolean deletePlugin(String pluginId) {
        checkPluginId(pluginId);

        PluginWrapper pluginWrapper = getPlugin(pluginId);
        PluginState pluginState = stopPlugin(pluginId);
        if (PluginState.STARTED == pluginState) {
            log.error("Failed to stop plugin '{}' on delete", pluginId);
            return false;
        }

        if (!unloadPlugin(pluginId)) {
            log.error("Failed to unload plugin '{}' on delete", pluginId);
            return false;
        }

        Path pluginPath = pluginWrapper.getPluginPath();

        return pluginRepository.deletePluginPath(pluginPath);
    }
```
The method check the plugin-state and try to unload it.
So why not add a plugin state named ```UNLOADED``` to indicate the plugin can be remmoved and after the method ```unloadPlugin``` called? 

The ```deletePlugin``` can be simple if the new state added.
```java
    @Override
    public boolean deletePlugin(String pluginId) {
        checkPluginId(pluginId);

        PluginWrapper pluginWrapper = getPlugin(pluginId);
        PluginState pluginState = pluginWrapper.getPluginState();
        if (PluginState.UNLOADED == pluginState) {
            Path pluginPath = pluginWrapper.getPluginPath();
            return pluginRepository.deletePluginPath(pluginPath);
        }
        log.error("Plugin '{}' not unload, can't be deleted", pluginId);
        return false;
    }
```
And it can indicate the plugin will not be invoked.
And users also can remove the plugin file by system command.

slovdahl:
Fixes #198.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/PluginState.java
+++ b/pf4j/src/main/java/org/pf4j/PluginState.java
@@ -20,7 +20,7 @@ package org.pf4j;
  * <p>
  * Lifecycle of a plugin:
  * <pre>
- * CREATED -> RESOLVED -> STARTED -> STOPPED
+ * CREATED -> RESOLVED -> STARTED -> STOPPED -> UNLOADED
  * CREATED -> DISABLED
  * CREATED -> FAILED
  *
@@ -57,7 +57,14 @@ public enum PluginState {
     /**
      * Plugin failed to start.
      */
-    FAILED("FAILED");
+    FAILED("FAILED"),
+
+    /**
+     * The plugin has been unloaded. After this event has been completed, the plugin's
+     * {@link ClassLoader} will be closed.
+     */
+    UNLOADED("UNLOADED"),
+    ;
 
     private final String status;
 
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -311,6 +311,7 @@ public abstract class AbstractPluginManager implements PluginManager {
         }
 
         // remove the plugin
+        pluginWrapper.setPluginState(PluginState.UNLOADED);
         plugins.remove(pluginId);
         getResolvedPlugins().remove(pluginWrapper);
         getUnresolvedPlugins().remove(pluginWrapper);
--- a/pf4j/src/test/java/org/pf4j/DefaultPluginManagerTest.java
+++ b/pf4j/src/test/java/org/pf4j/DefaultPluginManagerTest.java
@@ -25,6 +25,9 @@ import org.pf4j.test.PluginZip;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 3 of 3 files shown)


pf4j/src/main/java/org/pf4j/PluginState.java
```diff
@@ -57,7 +57,14 @@ public enum PluginState {
     /**
      * Plugin failed to start.
      */
-    FAILED("FAILED");
+    FAILED("FAILED"),
+
+    /**
+     * The plugin has been unloaded. After this event has been completed, the plugin's
+     * {@link ClassLoader} will be closed.
```
slovdahl, line 64: I think the second sentence here is important. This even needs to be posted _before_ closing the `ClassLoader`, otherwise we prevent event listeners from doing some things that involve the classes in the plugin itself. I'm just not sure how to write a test for it :thinking: Do you have any ideas @decebals?

slovdahl:
I'll rebase it tomorrow to get the conflicts resolved.

pf4j/src/main/java/org/pf4j/PluginState.java
decebals: I think that we can add a flag "closed" in `PluginClassLoader` and a method `isClosed():boolean`.  In the test method we will test the value returned by the `isClosed` method. In the future maybe we can use this flag to throw some exceptions in different methods with the error message "The class loader is already closed".
Or we can skip this test for the moment.
PluginClassLoader extends URLClassLoader and URLCLassLoader contains a [close](https://docs.oracle.com/javase/8/docs/api/java/net/URLClassLoader.html#close--) method with the below javadoc:

> Closes this URLClassLoader, so that it can no longer be used to load new classes or resources that are defined by this loader.

We need to investigate in the future more carefully this part with closed ClassLoader and the implications of this operation.

slovdahl: Good idea :+1: I gave it a try, pushed it as a separate commit.

> We need to investigate in the future more carefully this part with closed ClassLoader and the implications of this operation.

This would not hurt. I don't know if it's a problem, but I think we are forced to make the `boolean closed` `PluginClassLoader` field `volatile`, so checking it in the overridden method comes at some cost. OTOH, the methods are already quite heavy, and should maybe not be called on super-hot paths.

It could also cause some inconsistencies, unless we override more of the methods from the superclass :thinking: But that is probably just a minor annoyance.

perlun:
Nice to see this getting implemented, thanks @slovdahl and @decebals for good cooperation on this. :slightly_smiling_face:

PR #570, opened 2024-03-11
Avoid instantiating Plugin class if a plugin is already unloaded

slovdahl:
Given the following preconditions:
- `PluginClassLoader` is configured to use `ClassLoadingStrategy `PDA`, `DPA` or `DAP`.
- There is a `PluginStateListener` that calls `event.getPlugin().getPlugin()`.
- A plugin (`plugin-a`) with a missing dependency is loaded (`plugin-a`).
- The `Plugin` class in `plugin-a` extends, implements or references a class in either `plugin-b` or in the main application.

In that scenario, when `plugin-a` is unloaded by pf4j, and the `PluginStateListener` is run, the `event.getPlugin().getPlugin()` call will fail with a `NullPointerException`. This is caused by the class loading done when instantiating the `Plugin` class in `plugin-a`. Because not all classes needed are found in `plugin-a` itself, pf4j will continue by looking up classes in `plugin-a`'s dependencies. Because `plugin-b` is missing and has no associated `ClassLoader`.


pf4j/src/main/java/org/pf4j/PluginWrapper.java
```diff
@@ -90,7 +90,7 @@ public ClassLoader getPluginClassLoader() {
      * @return the plugin instance
      */
     public Plugin getPlugin() {
-        if (plugin == null) {
+        if (plugin == null && pluginState != PluginState.UNLOADED) {
```
slovdahl, line 93: @decebals thoughts about this? Does it make any sense at all? :thinking: 

We have some internal tests that use custom-built plugins, and now that I was rebuilding them to use a base `org.pf4j.Plugin` class coming from the application itself, those tests started failing.

The NPE is coming from the `PluginStateListener` that tries to tear down the Spring application context for unloaded plugins. We can work around it there too, but I figured it could make sense to avoid this in pf4j itself too.

In our `PluginStateListener` we did something like this before:
```java
if (event.getPluginState() == PluginState.UNLOADED) {
  if (pluginWrapper.getPlugin() instanceof SpringPlugin springPlugin) {
    ApplicationContext context = springPlugin.getPluginApplicationContext();
    if (childApplicationContext instanceof ConfigurableApplicationContext) {
      ((ConfigurableApplicationContext)context).close();
    }
  }
}
```

The workaround is to only do it if the previous state was `PluginState.STOPPED`:

```java
if (event.getPluginState() == PluginState.UNLOADED && event.getOldState() == PluginState.STOPPED) {
  if (pluginWrapper.getPlugin() instanceof SpringPlugin springPlugin) {
    ApplicationContext context = springPlugin.getPluginApplicationContext();
    if (childApplicationContext instanceof ConfigurableApplicationContext) {
      ((ConfigurableApplicationContext)context).close();
    }
  }
}
```

I would like to write some tests for this in pf4j too, but it requires a bit of new infrastructure for being able to create plugins that actually contain fully working classes.

decebals:
@slovdahl Is it still current?

slovdahl:
> @slovdahl Is it still current?

Need to double-check on Wednesday when I'm back at work, but I'm pretty sure we still have the workaround in place. So I think it still makes sense.

slovdahl:
Yes, we do have it in place, and we're running 3.12.0. I rebased the PR on top of latest master now.

pf4j/src/main/java/org/pf4j/PluginWrapper.java
decebals: After some analysis I came to the conclusion that this modification can suppress a possible business logic in certain projects.
I propose a simple solution that involves adding a new method in the `PluginWrapper`:
```
/**
  * Returns an {@link Optional} containing the plugin instance if it has been created.
  *
  * @return an Optional containing the plugin instance if created, otherwise an empty Optional
  */
 public Optional<Plugin> getPluginIfCreated() {
     return Optional.ofNullable(plugin);
 }
```
Using this method you can find out if a Plugin instance has already been created or not.
What do you think, solve your scenario?

decebals: @slovdahl What you say about my previous comment?

slovdahl: Sorry for the delay. I'll get back to you later this week (working 60% right now).

slovdahl: > Using this method you can find out if a Plugin instance has already been created or not.
> What do you think, solve your scenario?

Thanks for the analysis. Yep, I think this can also work for us. Instead of checking `event.getPluginState() == PluginState.UNLOADED && event.getOldState() == PluginState.STOPPED` we can do `event.getPluginState() == PluginState.UNLOADED && event.getPluginIfCreated().isPresent()`.

> After some analysis I came to the conclusion that this modification can suppress a possible business logic in certain projects.

Just out of curiosity, in which cases would it be warranted to create the `Plugin` if the current state is `UNLOADED`?

decebals: > Just out of curiosity, in which cases would it be warranted to create the Plugin if the current state is UNLOADED?

I don't have a clear case in my mind, I talk in general in terms of compatibility.
Probably an idea is to prevent (throw an exception) the plugin creation if the state is UNLOADED, but I'm afraid to add constraints.

Issue #574, opened 2024-03-27
v3.11.0 plugin status can‘t change from STARTED to DISABLED.


bigtrouble:
Looks like a bug （ it is very well in v3.10.0 )
I think there is something wrong with the code below.
`org.pf4j.AbstractPluginManager#disablePlugin` `L530`
```java
    @Override
    public boolean disablePlugin(String pluginId) {
        checkPluginId(pluginId);

        PluginWrapper pluginWrapper = getPlugin(pluginId);
        PluginDescriptor pluginDescriptor = pluginWrapper.getDescriptor();
        PluginState pluginState = pluginWrapper.getPluginState();
        if (PluginState.DISABLED == pluginState) {
            log.debug("Already disabled plugin '{}'", getPluginLabel(pluginDescriptor));
            return true;
        } else if (PluginState.STARTED == pluginState) {   
            if (PluginState.STOPPED == stopPlugin(pluginId)) {   <---------------------  !=  ????
                log.error("Failed to stop plugin '{}' on disable", getPluginLabel(pluginDescriptor));
                return false;
            }
        }
```

decebals:
I will investigate.
A unit test that highlighted the problem simplifies the investigation.

JohnNiang:
Hi @decebals , I'm willing to fix the problem.

PR #577, opened 2024-04-03
Fix the problem of not being able to disable plugins correctly

JohnNiang:
Closes https://github.com/pf4j/pf4j/issues/574

The problem might be introduced in <https://github.com/pf4j/pf4j/commit/30f415fe2d92d38ab84cac119e52fd54a4bc36ba>.


decebals:
@JohnNiang Thanks!

PR #579, opened 2024-04-06
Fix #576

### LINKED ISSUE #576: Error due to incorrect dependency update after plugin deletion
### Description
While managing plugins with p4fj, I encountered an issue where the dependency relationship of a deleted plugin was not correctly updated. Specifically, after deleting a plugin that other plugins depend on, attempting to stop those dependent plugins results in an error, stating `Plugin 'hello-plugin' not found`, despite `hello-plugin` having been deleted.

### Steps to Reproduce
1. Assume there are two plugins: one named `welcome-plugin` and another named `hello-plugin`, with `hello-plugin` depending on `welcome-plugin`.
2. Start all plugins.
3. Stop and delete the `hello-plugin`:
   ```java
   pluginManager.stopPlugin("hello-plugin");
   pluginManager.deletePlugin("hello-plugin");
   ```
4. Then, stop the `welcome-plugin`.

### Expected Behavior
After `hello-plugin` has been deleted, `welcome-plugin` should stop without any errors, unaffected by any dependency issues.

### Actual Behavior
When attempting to stop `welcome-plugin`, the error `Plugin 'hello-plugin' not found` occurs. This error is due to the plugin manager's `dependencyResolver` still referencing the uninstalled `hello-plugin`.

### Dependency Graphs
- `dependenciesGraph`:
   ```
   welcome-plugin -> []
   hello-plugin -> [welcome-plugin]
   ```
- `dependentsGraph`:
   ```
   welcome-plugin -> [hello-plugin]
   hello-plugin -> []
   ```

### Possible Solution
Update the dependency graphs upon plugin deletion to ensure all dependencies and dependents are correctly handled.

### Environment Information

v3.11.0

If I have misunderstood, please correct me 🫡

decebals:


pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
```diff
@@ -331,6 +331,9 @@ protected boolean unloadPlugin(String pluginId, boolean unloadDependents) {
             }
         }
 
+        // resolve the plugins again (update plugins graph)
+        resolveDependencies();
```
decebals, line 335: Here, I have some uncertainties regarding performance (the case in which a recursive upload is made => resolveDependencies method is called several times)

guqing: @decebals Can we update the dependencies graph by providing a method similar to `void removePlugin (PluginDescriptor descriptor)` in DependencyResolver

decebals: @guqing I know that I tried to do this, but it is not trivial, I have encountered more complex scenarios. I have some modifications on a local branch.
By the way, did you have the opportunity to test the fix from this PR? It is not optimal in terms of performance, but I don't think that performance is a problem in the given scenario.

guqing: Hi @decebals ，After testing it on my application, I found it works very well. I appreciate you taking the time to address this issue ♥️

decebals: @guqing I added a new commit to this PR that eliminate the possible performance penalty.
Please check this PR in this form (with this latest commit). If it's OK I will release a new version that includes this fix.

guqing: Hi @decebals , testing with the latest commit still works very well 👍

PR #595, opened 2024-11-11
Fix StackOverFlow error on Kotlin classes without @Extension

DRSchlaubi:
Currently this method gets caught in a loop (target -> retention -> mustbedocumented -> target) if @Extension is not present

this fixes it


decebals:
A (unit) test is welcome.

DRSchlaubi:
Would appriciate if we could get a release with this

PR #599, opened 2024-12-04
Fix the deprecation warnings from GitHub Actions

JohnNiang:
This PR fixes the deprecate warnings from [GitHub Actions](https://github.com/pf4j/pf4j/actions/runs/12144934668).

See
- https://github.blog/changelog/2024-09-16-notice-of-upcoming-deprecations-and-changes-in-github-actions-services/
- https://github.com/actions/runner-images/issues/10636

for more.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,9 +8,9 @@ on:
 jobs:
     build:
         name: Build
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-24.04
         steps:
-            - uses: actions/checkout@v2
+            - uses: actions/checkout@v4
               with:
                   fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
             - name: Set up JDK 17
@@ -20,11 +20,10 @@ jobs:
                   java-version: 17
                   cache: maven
             - name: Cache SonarCloud packages
-              uses: actions/cache@v1
+              uses: actions/cache@v4
               with:
                   path: ~/.sonar/cache
                   key: ${{ runner.os }}-sonar
-                  restore-keys: ${{ runner.os }}-sonar
             - name: Build and analyze
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any


.github/workflows/build.yml
```diff
@@ -20,11 +20,10 @@ jobs:
                   java-version: 17
                   cache: maven
             - name: Cache SonarCloud packages
-              uses: actions/cache@v1
+              uses: actions/cache@v4
               with:
                   path: ~/.sonar/cache
                   key: ${{ runner.os }}-sonar
-                  restore-keys: ${{ runner.os }}-sonar
```
decebals, line 27: Not clear why you are proposing to remove this line? Do you have a reason?

.github/workflows/build.yml
```diff
@@ -8,9 +8,9 @@ on:
 jobs:
     build:
         name: Build
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-24.04
```
decebals, line 11: I don't think we need this modification.
From what I understood the `ubuntu-latest` version is now mapped to 24.04. In the future will be other version, we want to stay with the latest version.

decebals (changes requested):

.github/workflows/build.yml
JohnNiang: I found the `key` and `restore-keys` are the same, and we could omit the `restore-keys` config according to <https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#using-the-cache-action>.

.github/workflows/build.yml
JohnNiang: I see your consideration. 

But the `latest` runner might bring some break changes that stop our workflow. Some actions might not be compatible with the `latest` runner, for example, <https://github.com/actions/runner-images/issues/10788>.  Using a runner with specific version is more reasonable than `latest`.

If we want to keep with the `latest` runner, we can migrate it manually to make sure all actions work very well.

JohnNiang (commented):
Hi @decebals , thank you for reviewing!

I left two replies, please take a moment to review.

PR #625, opened 2026-01-06
Fix exception handling inconsistency in startPlugin()

### LINKED ISSUE #624: Exception thrown when a plugin fails to start
First we compare the two code snippets below,

https://github.com/pf4j/pf4j/blob/6cd9a6cfd942f3b5c7017477172c7b489680bcfd/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L441-L443

https://github.com/pf4j/pf4j/blob/6cd9a6cfd942f3b5c7017477172c7b489680bcfd/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L396-L399

When we start a `single` plugin, failed state won't be captured. But we capture the exception while starting all plugins.

IMO, we should capture the exception and set failed state to the plugin. BTW I'm willing to propose a PR.

decebals:
Fixes #624

Problem:
The startPlugin() method did not handle exceptions during plugin startup, while startPlugins() properly caught exceptions and set the plugin state to FAILED. This created inconsistent behavior.

Solution:
Refactored exception handling logic into a private doStartPlugin() method that is now used by both startPlugin() and startPlugins(). This ensures consistent exception handling across both methods.

Changes:
- Extracted doStartPlugin() method with proper try-catch-finally
- Both methods now set plugin state to FAILED on exception
- Both methods now store the exception in failedException
- Both methods now fire state change events correctly
- Added comprehensive tests for exception scenarios

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -387,19 +387,7 @@ public abstract class AbstractPluginManager implements PluginManager {
         for (PluginWrapper pluginWrapper : resolvedPlugins) {
             PluginState pluginState = pluginWrapper.getPluginState();
             if (!pluginState.isDisabled() && !pluginState.isStarted()) {
-                try {
-                    log.info("Start plugin '{}'", getPluginLabel(pluginWrapper.getDescriptor()));
-                    pluginWrapper.getPlugin().start();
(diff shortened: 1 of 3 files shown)


decebals:
@JohnNiang Can you take a look (code review)?

JohnNiang (commented):
Hi @decebals , thank you for your effort!

I left one comment below, please take a moment to review.

JohnNiang (approved):
LGTM

PR #629, opened 2026-01-07
Validate dependency state on plugin start

### LINKED ISSUE #628: Plugin can start with failed required dependencies
# Plugin can start with failed required dependencies

## Problem

When `startPlugin(pluginId)` starts a plugin's dependencies, it doesn't check if they actually started successfully. This means a plugin can end up in STARTED state even when its required dependencies are in FAILED state.

Example:
- Plugin A depends on Plugin B (required)
- Plugin B fails to start (throws exception)
- Plugin B state: FAILED
- Plugin A state: STARTED (wrong - should be FAILED)

This causes runtime errors when Plugin A tries to use Plugin B.

## Current Behavior

In `AbstractPluginManager.startPlugin()` around line 422:
```java
for (PluginDependency dependency : pluginDescriptor.getDependencies()) {
    if (!dependency.isOptional() || plugins.containsKey(dependency.getPluginId())) {
        startPlugin(dependency.getPluginId());  // Return value is ignored
    }
}
return doStartPlugin(pluginWrapper);  // Starts regardless
```

## Expected Behavior

**For required dependencies:**
- If dependency fails to start, the dependent plugin should also fail to start
- Clear error message indicating which dependency failed

**For optional dependencies:**
- If dependency fails to start, log a warning but allow the dependent plugin to start
- Plugin can run with reduced functionality

## Context

This issue was identified by @JohnNiang during review of PR #625. It's a pre-existing issue, not introduced by that PR.

## Open Question

What should happen if an **optional** dependency is present (loaded) but fails to start?

A. Let dependent plugin start with WARNING log (proposed)
B. Fail the dependent plugin too (stricter)

Feedback welcome.

## Notes

- This is a breaking change for required dependencies (correct behavior, but some plugins that currently start will fail)
- No tests currently exist for this scenario
- Will be fixed in a separate PR from #625

decebals:
Fixes #628

   ## Summary

   This PR adds validation of dependency state during plugin startup to prevent plugins from starting when their required dependencies have failed to start.

   ## Changes

   - Extract `startDependencies()` method with comprehensive JavaDoc documenting behavior for required vs optional dependencies
   - Validate dependency state after starting each dependency
   - Required dependencies that fail to start now prevent the dependent plugin from starting (plugin state set to FAILED)
   - Optional dependencies that fail to start log a warning but allow the dependent plugin to start (degraded mode)
   - Add tests for required dependency failures, optional dependency failures, and cascading failures

   ## Behavior

   **Required dependencies:**
   - If a required dependency fails to start, the dependent plugin will NOT start
   - Plugin state is set to FAILED with a clear error message
   - Prevents runtime errors by failing fast at startup

   **Optional dependencies:**
   - If not loaded, skipped (existing behavior)
   - If loaded but fails to start, logs a WARNING and allows dependent to start
   - Plugin is responsible for handling missing optional functionality gracefully

   ## Testing

   Added 3 new tests:
   - `shouldFailToStartPluginWhenRequiredDependencyFailsToStart` - verifies required dependency validation
   - `shouldStartPluginWhenOptionalDependencyFailsToStart` - verifies optional dependency handling
   - `shouldHandleCascadingDependencyFailures` - verifies cascading failure propagation

   All 26 tests pass.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -419,14 +419,77 @@ public abstract class AbstractPluginManager implements PluginManager {
             }
         }
 
-        for (PluginDependency dependency : pluginDescriptor.getDependencies()) {
-            // start dependency only if it marked as required (non-optional) or if it optional and loaded
-            if (!dependency.isOptional() || plugins.containsKey(dependency.getPluginId())) {
-                startPlugin(dependency.getPluginId());
-            }
+        // Start and validate dependencies
+        if (!startDependencies(pluginWrapper)) {
+            return PluginState.FAILED;
         }
 
         return doStartPlugin(pluginWrapper);
+    }
+
+    /**
+     * Starts all dependencies of a plugin and validates that required dependencies started successfully.
+     * <p>
+     * This method is called before {@link #doStartPlugin(PluginWrapper)} to ensure all dependencies
+     * are in a valid state before starting the dependent plugin.
+     * <p>
+     * <b>Behavior for required dependencies:</b>
+     * <ul>
+     *   <li>If a required dependency fails to start (state is not {@link PluginState#STARTED}),
+     *       the dependent plugin will NOT be started</li>
+     *   <li>The dependent plugin state is set to {@link PluginState#FAILED}</li>
+     *   <li>A {@link PluginRuntimeException} is set as the failure cause</li>
+     *   <li>An error is logged indicating which dependency failed</li>
+     *   <li>A state change event is fired</li>
+     * </ul>
+     * <p>
+     * <b>Behavior for optional dependencies:</b>
+     * <ul>
+     *   <li>If an optional dependency is not loaded, it is skipped (no attempt to start)</li>
+     *   <li>If an optional dependency is loaded but fails to start, a warning is logged</li>
+     *   <li>The dependent plugin will still be started (allows degraded mode/reduced functionality)</li>
+     *   <li>It is the plugin's responsibility to handle missing optional dependencies gracefully</li>
+     * </ul>
+     *
+     * @param pluginWrapper the plugin whose dependencies should be started
+     * @return {@code true} if all required dependencies started successfully and the plugin can proceed to start;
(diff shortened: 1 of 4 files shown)


decebals:
Implementation is complete and all tests pass (26/26).

Marked as draft to gather feedback on the approach before merging, particularly regarding the behavior for optional dependencies:

**Current implementation:**
- Required dependencies that fail → dependent plugin fails to start (FAILED state)
- Optional dependencies that fail → WARNING logged, dependent plugin starts anyway (degraded mode)

**Question:** Is this the right behavior for optional dependencies that are loaded but fail to start? Should they:
- A) Allow dependent to start with WARNING (current implementation)
- B) Prevent dependent from starting (stricter)
- C) Something else?

The rationale for (A) is that "optional" should mean the plugin can function without it, even if it's present but broken. But I'm open to other perspectives.

Ready to mark as ready for review once we have consensus on the approach.

decebals (commented):
Hi @JohnNiang! Thanks for the feedback.

I reviewed the current implementation and wanted to confirm what we already do when a plugin fails to start due to dependency issues:

**Current behavior (lines 476-486):**
```java
// Required dependency failed: fail fast
log.error("Cannot start plugin '{}' because required dependency '{}' failed to start (state: {})",
        pluginId, dependency.getPluginId(), dependencyState);

pluginWrapper.setPluginState(PluginState.FAILED);
pluginWrapper.setFailedException(
    new PluginRuntimeException("Required dependency '" + dependency.getPluginId() + "' failed to start")
);
firePluginStateEvent(new PluginStateEvent(this, pluginWrapper, pluginState));
return false;
```

So we already:
- Set FAILED state
- Set a failure exception with descriptive message
- Fire state change event

The caller can check `pluginWrapper.getFailedException()` to understand what happened.

**Question:** Are you suggesting we should create a **specific exception type** (like `PluginDependencyException extends PluginRuntimeException`) instead of using the generic `PluginRuntimeException`? This would allow callers to use `instanceof` checks rather than parsing the exception message.

If so, I think that's a good improvement for better exception handling! Would you like me to:
1. Create a new `PluginDependencyException` class with a `getDependencyId()` method
2. Update the implementation to use it
3. Add tests for the new exception type

Let me know your thoughts!

JohnNiang (approved):
LGTM

PR #631, opened 2026-01-08
Set failedException when plugin validation or stop fails (#630)

### LINKED ISSUE #630: Failure reason not set in PluginWrapper while failing to enable or start plugins
Currently, if we fail to enable or start plugins, only error or warn logs will be printed, such as 

https://github.com/pf4j/pf4j/blob/b815351110d29f80368e4ba5f9bd1b11c2daab76/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java#L819

Sometime, we have to enable or start plugins at runtime instead of application startup. Caller has no way of knowing what just happened, other than checking the logs.

It would be very useful if we could know the failure reason from `org.pf4j.PluginWrapper#failedException` when the `org.pf4j.PluginWrapper#pluginState` is `FAILED`.

decebals:
Fixes #630

  ## Problem

  When plugins fail to start or enable due to validation errors or exceptions, the PluginWrapper's failedException field is not always set. This makes it impossible for callers to programmatically determine why a plugin failed.

  ## Solution

  This PR ensures that failedException is consistently set in the following scenarios:

  1. **Plugin validation failure during load** - When a plugin doesn't meet system version requirements, failedException is now set with a generic message. The detailed reason remains in the warning log.

  2. **Exception during plugin stop** - When stopPlugins() catches an exception from plugin.stop(), the plugin is now marked as FAILED with the exception preserved. Previously, the exception was only logged and the plugin remained in STARTED state.

  3. **Validation failure during enable** - When enablePlugin() is called on an invalid plugin, failedException is now set before returning false.

  ## Test Coverage

  Added three comprehensive tests to verify the new behavior:
  - invalidPluginSetsFailedException
  - stopPluginsWithExceptionSetsFailedException
  - enableInvalidPluginSetsFailedException

  Also added PluginZip.Builder.pluginRequires() helper method and FailingStopPlugin test utility class.

### DIFF
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -538,6 +538,9 @@ public abstract class AbstractPluginManager implements PluginManager {
                     firePluginStateEvent(new PluginStateEvent(this, pluginWrapper, pluginState));
                 } catch (PluginRuntimeException e) {
                     log.error(e.getMessage(), e);
+                    pluginWrapper.setPluginState(PluginState.FAILED);
+                    pluginWrapper.setFailedException(e);
+                    firePluginStateEvent(new PluginStateEvent(this, pluginWrapper, pluginState));
                 }
             } else {
                 // do nothing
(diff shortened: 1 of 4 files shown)


decebals:
## Additional Changes

Added two more commits to improve exception handling consistency and test coverage:

**Commit 2: Extract doStopPlugin() method**

Following the same pattern as doStartPlugin(), created a private doStopPlugin() method that centralizes exception handling for all plugin stop operations. This ensures failedException is consistently set when stop() throws an exception in:
- stopPlugins()
- disablePlugin()  
- deletePlugin()
- unloadPlugin()

The refactoring eliminates manual exception checking in calling methods since doStopPlugin() automatically sets the FAILED state and captures the exception.

**Commit 3: Add coverage test**

Added unloadPluginWithStopFailureSetsFailedException test to increase coverage of the exception handling path in unloadPlugin() when stopPlugin() throws an exception.

Total test coverage: 6 tests covering all failure scenarios where failedException should be set.

JohnNiang (approved):
LGTM

PR #633, opened 2026-01-13
Allow customization of PluginClassLoader parent delegation

### LINKED ISSUE #632: Allow customization of parent delegation exclusions in PluginClassLoader
## Context
While developing PF4J-Plus (a lightweight toolkit providing platform-level services for PF4J applications), I encountered a limitation in `PluginClassLoader`.

## Problem
`PluginClassLoader.loadClass()` has hardcoded exclusions for plugin packages:

```java
if (className.startsWith(PLUGIN_PACKAGE_PREFIX) 
    && !className.startsWith("org.pf4j.demo") 
    && !className.startsWith("org.pf4j.test")) {
    return getParent().loadClass(className);
}
```

These exclusions (`org.pf4j.demo`, `org.pf4j.test`) are specific to PF4J's own demo/test structure and cannot be customized by framework extensions or users with different package naming conventions.

## Current Behavior
- PF4J-Plus demo plugins use package `org.pf4j.plus.demo.plugins.*`
- These classes start with `org.pf4j.` but are NOT excluded from parent delegation
- Result: `ClassNotFoundException` because plugin classes are delegated to parent ClassLoader instead of being loaded from the plugin JAR

## Expected Behavior
Users should be able to customize which packages are excluded from parent delegation, especially when:
- Building framework extensions (like PF4J-Plus)
- Using custom package naming conventions
- Implementing specific class loading policies

## Proposed Solution
Extract the exclusion logic into a protected method:

```java
protected boolean shouldDelegateToParent(String className) {
    return className.startsWith(PLUGIN_PACKAGE_PREFIX) 
        && !className.startsWith("org.pf4j.demo") 
        && !className.startsWith("org.pf4j.test");
}
```

Then in `loadClass()`:
```java
if (shouldDelegateToParent(className)) {
    return getParent().loadClass(className);
}
```

This allows easy customization through inheritance:
```java
@Override
protected boolean shouldDelegateToParent(String className) {
    return super.shouldDelegateToParent(className) 
        && !className.startsWith("org.pf4j.plus.demo");
}
```

## Benefits
- Backwards compatible (no API changes)
- Follows PF4J's design pattern (protected factory methods)
- Minimal code changes (just extract to method)
- Enables framework extensions and custom use cases
- No impact on existing users

## Use Case
PF4J-Plus is a companion library adding ServiceRegistry, dependency injection, and EventBus to PF4J applications. It follows PF4J conventions but needs `org.pf4j.plus.demo.*` excluded from parent delegation.

decebals:
## Summary
   Extracts hardcoded package exclusions from `PluginClassLoader` into a protected `shouldDelegateToParent()` method that can be overridden by subclasses.

   ## Problem
   Previously, the parent delegation logic was hardcoded to exclude `org.pf4j.demo` and `org.pf4j.test` packages. This prevented framework extensions (like PF4J-Plus) from customizing class loading behavior for
   their own demo or test packages.

   ## Solution
   - Add `protected boolean shouldDelegateToParent(String className)` method to `PluginClassLoader`
   - Remove `org.pf4j.demo` exclusion (not part of core framework)
   - Keep `org.pf4j.test` exclusion (part of core test utilities)
   - Provide working examples in both Maven and Gradle demos via custom `DemoPluginClassLoader` and `DemoPluginLoader` classes

   ## Changes
   - **PluginClassLoader.java**: Extract exclusion logic into overridable method
   - **PluginClassLoaderTest.java**: Add 5 new tests demonstrating override capability
   - **demo/gradle**: Add `DemoPluginClassLoader`, `DemoPluginLoader`; update to PF4J 3.15.0-SNAPSHOT
   - **demo/maven**: Add `DemoPluginClassLoader`, `DemoPluginLoader`

   ## Backwards Compatibility
   Fully backwards compatible. Follows existing pattern of protected factory methods used throughout the framework.

   ## Testing
   All tests pass (33/33 in PluginClassLoaderTest). New tests demonstrate:
   - Default behavior for PF4J core classes
   - Exclusions for test utilities
   - Override capability for custom packages (e.g., `org.pf4j.plus.demo`)

   Fixes #632

### DIFF
--- a/demo/gradle/app/src/main/java/org/pf4j/demo/DemoPluginLoader.java
+++ b/demo/gradle/app/src/main/java/org/pf4j/demo/DemoPluginLoader.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2012-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pf4j.demo;
+
+import org.pf4j.BasePluginLoader;
+import org.pf4j.DefaultPluginClasspath;
+import org.pf4j.PluginClassLoader;
+import org.pf4j.PluginDescriptor;
+import org.pf4j.PluginManager;
+
+import java.nio.file.Path;
+
+/**
+ * Custom PluginLoader that uses {@link DemoPluginClassLoader} for loading plugins.
+ *
+ * @author Decebal Suiu
+ */
+class DemoPluginLoader extends BasePluginLoader {
+
+    public DemoPluginLoader(PluginManager pluginManager) {
+        super(pluginManager, new DefaultPluginClasspath());
+    }
+
+    @Override
+    protected PluginClassLoader createPluginClassLoader(Path pluginPath, PluginDescriptor pluginDescriptor) {
+        return new DemoPluginClassLoader(pluginManager, pluginDescriptor, getClass().getClassLoader());
+    }
+
+}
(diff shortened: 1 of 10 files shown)


demo/gradle/app/src/main/java/org/pf4j/demo/DemoPluginLoader.java
```diff
@@ -0,0 +1,56 @@
...
+import java.nio.file.Path;
+
+/**
+ * Custom PluginLoader that uses {@link DemoPluginClassLoader} for loading plugins.
+ *
+ * @author Decebal Suiu
+ */
+class DemoPluginLoader extends CompoundPluginLoader {
+
+    public DemoPluginLoader(PluginManager pluginManager) {
```
decebals, line 35: I don't like this verbosity but for the moment I don't have a solution.

JohnNiang (approved):
LGTM

PR #644, opened 2026-01-21
Rename LegacyExtension* classes to IndexedExtension*

### LINKED ISSUE #643: Rename LegacyExtension* classes to IndexedExtension*
## Problem

The current naming of `LegacyExtensionFinder` and `LegacyExtensionStorage` is misleading:

- **\"Legacy\" implies deprecated/old** - something that should be avoided or replaced
- **In reality, these are the DEFAULT classes** used by PF4J (see `DefaultExtensionFinder` and `ExtensionAnnotationProcessor`)
- `ServiceProviderExtensionFinder` is actually commented out in `DefaultExtensionFinder`

This causes confusion for developers who might think they should avoid \"Legacy\" classes when in fact they're the standard mechanism.

## Proposal

Rename to `IndexedExtension*` which accurately describes the mechanism:
- Uses `META-INF/extensions.idx` - an **index file**
- `IndexedExtensionFinder` - finds extensions using the index file
- `IndexedExtensionStorage` - stores extensions in the index file

## Implementation (backwards compatible)

1. Create new classes with the correct names
2. Keep old classes as deprecated aliases:

```java
@Deprecated(since = "3.x", forRemoval = true)
public class LegacyExtensionFinder extends IndexedExtensionFinder {}
```

3. Remove aliases in next major version (4.0)

## Impact

Low - these are internal classes used only by platform/app developers who customize extension finding. Plugin developers never touch them.

Related: #207

decebals:
## Summary                                                                                                                                                                                                                          
  - Rename `LegacyExtensionFinder` → `IndexedExtensionFinder`                                                                                                                                                                         
  - Rename `LegacyExtensionStorage` → `IndexedExtensionStorage`                                                                                                                                                                       
  - Legacy classes become deprecated aliases for backwards compatibility                                                                                                                                                              
                                                                                                                                                                                                                                      
  ## Motivation                                                                                                                                                                                                                       
  "Legacy" naming was misleading - these are the **default** classes, not deprecated/old code. "Indexed" accurately describes the mechanism (uses `META-INF/extensions.idx`).                                                         
                                                                                                                                                                                                                                      
  ## Test plan                                                                                                                                                                                                                        
  - [x] All existing tests pass                                                                                                                                                                                                       
  - [x] Build compiles successfully                                                                                                                                                                                                   
                                                                                                                                                                                                                                      
  Fixes #643


decebals:
Closing to redo with `git mv` - this preserves Git history on the renamed files. With the current approach, `IndexedExtensionFinder.java` has no history; with `git mv`, `git log --follow` will show the complete history from when it was `LegacyExtensionFinder.java`.

PR #661, opened 2026-08-25
Fail the build when the SonarCloud quality gate fails

### LINKED ISSUE #659: Build should fail when the SonarCloud quality gate fails
mvn sonar:sonar sends the report and moves on. The build stays green even when the gate is red, so at the moment the gate is a dashboard rather than a gate.

`-Dsonar.qualitygate.wait=true` makes the build wait for the analysis to finish and fail if the gate does not pass.

Worth being clear about what this enforces. The default gate is Clean as You Code, which looks at new code only: no new issues, no new duplication, coverage on the lines a change touches. Existing gaps are untouched, so it should not block unrelated work.

decebals:
Fixes #659

Adds `-Dsonar.qualitygate.wait=true` so the build waits for the analysis to finish and fails if the gate does not pass.

One thing to sort out before merging: the gate currently reports no status at all on `master`.

```
$ curl -s "https://sonarcloud.io/api/qualitygates/project_status?projectKey=pf4j_pf4j&branch=master"
{"projectStatus":{"status":"NONE","conditions":[],"periods":[]}}
```

The Sonar way gate is assigned and analyses are running, so the likely cause is that no New Code period is defined for the project. Every condition in that gate measures new code, so with no baseline there is nothing to evaluate and the flag would be a no-op. Needs setting under Project Settings, New Code, where "Previous version" fits a library like this.

This PR is also the first test of the change, since it runs analysis in pull request mode.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
                   # SONAR_TOKEN is not exposed to pull requests from forks, and the
                   # scanner fails hard without it. Build without analysis in that case.
                   if [ -n "$SONAR_TOKEN" ]; then
-                      cd pf4j && mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
+                      cd pf4j && mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.qualitygate.wait=true
                   else
                       echo "SONAR_TOKEN unavailable (fork pull request). Skipping analysis."
                       cd pf4j && mvn -B verify


decebals:
The flag works. From this PR's build:

```
[INFO] ------------- Check Quality Gate status
[INFO] Waiting for the analysis report to be processed (max 300s)
[INFO] QUALITY GATE STATUS: PASSED
```

Correcting what I wrote above about New Code: it is not a prerequisite here. Pull request analysis uses the diff against the base branch as its baseline, so the gate evaluates without any New Code period being configured. The `NONE` status only applies to the `master` branch verdict, which matters far less, since the point is to block before merge rather than report after.

Analysis added about 5 seconds to the build.

PR #681, opened 2026-09-02
Document PluginManager.getExtensions instantiation #293

### LINKED ISSUE #293: Document how Extensions are instantiated
It would be nice if `org.pf4j.PluginManager#getExtensions` had API documentation.

Some things that might be useful to know (or know that they are unspecified, or implementation defined):
1. When extensions are instantiated (When the plugin is loaded/started? When explicitly requested)? (see #232)
2. In which plugin states is it OK to call this method? (In any? In loaded/started only? What happens in forbidden states?)
3. If I call this method several times and get extensions, is it guaranteed that I will get the same instances?
4. What will happen if I pass unknown pluginId or an absent extension type? (empty list/exception?)

AzazelSensei:
The getExtensions methods on PluginManager had almost no API docs.

I documented the current DefaultPluginManager / AbstractExtensionFinder behavior:

- instances are created on getExtensions, not at load/start
- only STARTED plugins contribute (null pluginId is the application classpath)
- DefaultExtensionFactory makes a new instance each call; SingletonExtensionFactory reuses
- unknown plugin id or missing type returns an empty list

Fixes #293

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [#673]: Read the extensions index of a plugin from every jar on the plugin classpath
 
 #### Added
+- [#293]: Document when `PluginManager.getExtensions` creates instances and what it returns
 
 #### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -653,6 +653,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 [#297]: https://github.com/pf4j/pf4j/issues/297
 [#296]: https://github.com/pf4j/pf4j/issues/296
 [#294]: https://github.com/pf4j/pf4j/issues/294
+[#293]: https://github.com/pf4j/pf4j/issues/293
 [#292]: https://github.com/pf4j/pf4j/issues/292
 [#288]: https://github.com/pf4j/pf4j/pull/288
 [#287]: https://github.com/pf4j/pf4j/pull/287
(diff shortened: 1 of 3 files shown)


decebals:
Thanks, this closes a gap that has been open since 2019, and the behaviour you describe matches the code. I checked each claim against `AbstractExtensionFinder` and `AbstractPluginManager`: the wrappers are rebuilt on every `find`, a plugin that is not started contributes nothing, a null plugin id reads the application classpath, and an instantiation failure is wrapped in a `PluginRuntimeException` that `getExtensions` logs and skips. Javadoc builds clean with the new links.

One thing to fix before this can go in: the changelog entry uses `[#293]` but the link definition at the bottom of the file is missing, so it renders as plain text. Please add

```
[#293]: https://github.com/pf4j/pf4j/issues/293
```

with the others, they are kept in descending order.

decebals:
Following up on my own review, I looked at what the claims say and not at where they sit, and I think part of this belongs one layer down.

`PluginManager` does not reference an implementation anywhere today, and this would be the first place naming `DefaultPluginManager`, `DefaultExtensionFactory` and `SingletonExtensionFactory`. An interface also cannot promise that something gets logged.

Could you keep on the interface what any implementation has to honour: never null, an empty list for an unknown id or a type with no extensions, a null plugin id meaning the application classpath, and that instances come from the `ExtensionFactory`, which decides whether they are shared or created per call.

The rest, that instances are created on the call rather than at load or start, that only started plugins contribute, the two factories, and the logging, reads well on the `AbstractPluginManager` overrides, where that behaviour actually lives. #293 asks for exactly that distinction, it accepts "implementation defined" as an answer.

If you would rather not come back to this, say so and we will do the move in a follow-up, the content is yours either way.

Sorry for coming back with this after approving the substance.

AzazelSensei:
Moved the DefaultPluginManager / factory / logging bits onto the AbstractPluginManager overrides and left the interface at the contract you listed (never null, empty list, null plugin id, ExtensionFactory identity). Added the [#293] changelog link.

decebals:
Thanks for the contribution, and for the quick turnaround. This is what I meant, and the changelog link is in the right place.

PR #686, opened 2026-09-02
Document when PluginManager.unloadPlugin returns false #291

### LINKED ISSUE #291: Document when unloadPlugin can fail
[`PluginManager#unloadPlugin`](https://static.javadoc.io/org.pf4j/pf4j/2.6.0/org/pf4j/PluginManager.html#unloadPlugin\(java.lang.String\)) _may_ return false to notify that it failed to unload the plugin. However, the documentation does not specify:
* Under which circumstances it can fail to unload the plugin (or can it really?).
* What the client code can do in this case.

Would it be possible to clarify that in the API docs please?

AzazelSensei:
unloadPlugin's javadoc only said it returns true if the plugin was unloaded.

I documented the current AbstractPluginManager behavior:

- false when the id is unknown or already unloaded (not an error)
- false when stop leaves the plugin STARTED; it stays loaded
- Plugin.stop throwing PluginRuntimeException still unloads and returns true; the cause is on PluginWrapper.getFailedException()
- PluginRuntimeException is thrown if the plugin classloader cannot be closed

Fixes #291

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Added
 - [#293]: Document when `PluginManager.getExtensions` creates instances and what it returns
+- [#291]: Document when `PluginManager.unloadPlugin` returns `false` and what callers can do
 
 #### Removed
 
@@ -657,6 +658,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 [#294]: https://github.com/pf4j/pf4j/issues/294
 [#293]: https://github.com/pf4j/pf4j/issues/293
 [#292]: https://github.com/pf4j/pf4j/issues/292
+[#291]: https://github.com/pf4j/pf4j/issues/291
 [#288]: https://github.com/pf4j/pf4j/pull/288
 [#287]: https://github.com/pf4j/pf4j/pull/287
 [#278]: https://github.com/pf4j/pf4j/pull/278
--- a/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
+++ b/pf4j/src/main/java/org/pf4j/AbstractPluginManager.java
@@ -263,10 +263,7 @@ public abstract class AbstractPluginManager implements PluginManager {
     }
 
     /**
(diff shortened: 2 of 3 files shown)


decebals:
Thanks for picking this one up, it has been open since 2019 and the behaviour you describe is real. I checked it against `AbstractPluginManager.unloadPlugin`. Four changes and this can go in.

**1. Split the text the way we did on #681.** The point there was not only to keep implementation class names off the interface, it was that `PluginManager` states what every implementation must honour and the behaviour of one implementation goes on the `AbstractPluginManager` override.

Keep on the interface: the plugin and its dependents are unloaded, dependents first; `true` when the plugin was unloaded; `false` when there is no such loaded plugin, or when it could not be stopped and therefore stays loaded.

Move to the override: that a `PluginRuntimeException` from `Plugin.stop()` does not prevent the unload, that the plugin is removed anyway with state `FAILED` and the cause on `PluginWrapper.getFailedException()`, and that the class loader is closed at the end. A javadoc body consisting only of `{@inheritDoc}` is the same as no comment, so that is the place for it.

**2. Drop "treat it as a no-op".** Dependents are unloaded before the `plugins.containsKey(pluginId)` check and before the stop, so `false` can come back after the dependents are already gone. Please say that instead.

**3. Move the `getFailedException()` pointer.** In the branch that returns `false` because the plugin is still `STARTED`, nothing sets the failed exception. `doStopPlugin` sets it only when `stop()` throws, and then the state is `FAILED` and the method returns `true`. So that pointer belongs to the successful unload, in the override. Worth adding there that the shipped stop path never leaves the plugin `STARTED`, since `doStopPlugin` returns `STOPPED` or `FAILED` and any other exception is caught and the plugin removed anyway; the branch is there for a subclass that overrides the stop.

**4. Two small ones.** Widen the `@throws`: `resolveDependencies()` runs after the removal and can throw `CyclicDependencyException` or `DependenciesNotFoundException`, both `PluginRuntimeException`. And in the changelog, entries under a heading go in ascending issue order, so `[#291]` before `[#293]`; the link definition at the bottom is already in the right place.

